PR #3, opened 2016-10-04
Update LRU when index is used

rnewson:
Prior to this commit only index opening/lookup would update the
LRU. This commit updates the LRU on index activity, as it should
always have done. The effect is that we will close the least-recently
used index from now on.

BugzID: 75594


mayya-sharipova:
+1

PR #7, opened 2016-11-30
Support cluster purge

chguocloudant:
- add 2 APIs(set_purge_seq, get_purge_seq).
- add a test case to test the new APIs.

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/IndexService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexService.scala
@@ -71,7 +71,6 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
   var updateSeq = getCommittedSeq
   var pendingSeq = updateSeq
   var purgeSeq = getCommittedPurgeSeq
-  var pendingPurgeSeq = purgeSeq
   var committing = false
   var forceRefresh = false
 
@@ -121,10 +120,9 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
         pendingSeq = newSeq
         logger.debug("Pending sequence is now %d".format(newSeq))
         'ok
-      case SetPurgeSeqMsg(newSeq: Long) =>
-        pendingPurgeSeq = newSeq
-        logger.debug("Pending purge sequence is now %d".format(newSeq))
-        commit(pendingSeq, pendingPurgeSeq)
+      case SetPurgeSeqMsg(newPurgeSeq: Long) =>
+        purgeSeq = newPurgeSeq
+        logger.debug("purge sequence is now %d".format(newPurgeSeq))
         'ok
       case 'info =>
         ('ok, getInfo)
@@ -157,13 +155,12 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
       }
       exit('deleted)
     case 'maybe_commit =>
-      commit(pendingSeq, pendingPurgeSeq)
-    case ('committed, newUpdateSeq: Long, newPurgeSeq: Long) =>
+      commit(pendingSeq, purgeSeq)
+    case ('committed, newUpdateSeq: Long) =>
       updateSeq = newUpdateSeq
-      purgeSeq = newPurgeSeq
       forceRefresh = true
       committing = false
-      logger.debug("Committed sequence %d and %d".format(newUpdateSeq, newPurgeSeq))
+      logger.debug("Committed sequence %d".format(newUpdateSeq))
     case 'commit_failed =>
       committing = false
   }
@@ -197,7 +194,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
           commitTimer.time {
             ctx.args.writer.commit()
           }
-          index ! ('committed, newUpdateSeq, newPurgeSeq)
+          index ! ('committed, newUpdateSeq)
         } catch {
           case e: AlreadyClosedException =>
             logger.error("Commit failed to closed writer", e)
@@ -581,7 +578,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
       ('doc_del_count, reader.numDeletedDocs),
       ('pending_seq, pendingSeq),
       ('committed_seq, getCommittedSeq),
-      ('purge_seq, pendingPurgeSeq)
+      ('purge_seq, purgeSeq)
     )
   }
 
--- a/src/main/scala/com/cloudant/clouseau/IndexService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexService.scala
@@ -70,6 +70,8 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
   var reader = DirectoryReader.open(ctx.args.writer, true)
   var updateSeq = getCommittedSeq
   var pendingSeq = updateSeq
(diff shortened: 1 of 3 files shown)


rnewson:
where is purgeSeq persisted to the index?

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -70,6 +70,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
   var reader = DirectoryReader.open(ctx.args.writer, true)
   var updateSeq = getCommittedSeq
   var pendingSeq = updateSeq
+  var purgeSeq = 0L
```
mayya-sharipova: why the default value is 0? Similarly to `updateSeq` it should be read from commitData on disk

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -199,6 +206,26 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
     }
   }
 
+  private def commitPurgeSeq(newSeq: Long) {
```
rnewson: This method does not set the committing flag and so could allow concurrent attempts to commit, leading to a broken index (one with the wrong values).

This work should instead be done in the preceding `commit` function.

chguocloudant: You mean we can commit `purgeSeq` and `updateSeq` with `commit` function?
We can change `commit` function with 2 arguments to support it.
```
  private def commit(newSeq: Long, type: String) {
    if (type == "updateSeq") {
      ...
    } else {
      ...
    }
```

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -569,7 +580,8 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
       ('doc_count, reader.numDocs),
       ('doc_del_count, reader.numDeletedDocs),
       ('pending_seq, pendingSeq),
-      ('committed_seq, getCommittedSeq)
+      ('committed_seq, getCommittedSeq),
+      ('purge_seq, pendingPurgeSeq)
```
rnewson: should be `purgeSeq` here not `pendingPurgeSeq`.

rnewson (approved):
Approved, nicely done. Please don't merge until we have the matching dreyfus work at the same level in case it changes what we need here.

src/main/scala/com/cloudant/clouseau/IndexService.scala
chguocloudant: OK, I will change it later.

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -117,6 +121,11 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
         pendingSeq = newSeq
         logger.debug("Pending sequence is now %d".format(newSeq))
         'ok
+      case SetPurgeSeqMsg(newSeq: Long) =>
+        pendingPurgeSeq = newSeq
+        logger.debug("Pending purge sequence is now %d".format(newSeq))
+        commit(pendingSeq, pendingPurgeSeq)
```
chguocloudant: Hi @rnewson .
I triggered commit when the purgeSeq is updated. Will it cost too much?

chguocloudant: And I think the var `pendingPurgeSeq` can be removed if the `commit` is allowed. Right?

rnewson: hm, good question. I'd rather we always commit asynchronously and on the established interval. For users that purge a lot (and there will be a lot of them if this work is extended to allow 'hard' deletes) they'll be committing very frequently and this will hurt performance.

This means the purge seq message could be processed but not actioned, when clouseau crashes before committing, so we'll need to ensure that dreyfus will retry (it should, since it reads the purge seq from the index, but let's be sure).

chguocloudant: Hi @rnewson .
If we commit purgeSeq asynchronously and also the users purge a lot, the `dreyfus` will retry to delete the index again in `clouseau`.
For example, if `dreyfus` gets a purgeSeq(6) from couchdb and gets an IdxPurgeSeq(5) from `clouseau`, it will purge index from 5 and set purgeSeq to 6.
If `clouseau` is crashed before committing, `dreyfus` will get purgeSeq(5) again. And `dreyfus` will purge index from 5 again.
But I am not sure which is more expensive, delete index or commit?

rnewson: we can safely assume clouseau crashes are rare from our production experience.

commit is expensive, the most expensive operation there is in Lucene, I think.

So the choice is between an expensive commit for every purge or a less expensive delete on retry for every clouseau crash.

We should also not say "delete index" when we really mean "delete document". :)

chguocloudant: OK. I think I can modify it later.
Let me make sure that we can remove the value `pendingPurgeSeq`, right? If the `dreyfus` calls `get_purge_seq`, `clouseau` will return `purgeSeq` directly(now this value maybe is not committed yet).

rnewson: I don't see how we can remove pending. we need to return the committed purge seq when asked and need to retain the pending value for when we commit.

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -117,6 +120,10 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
         pendingSeq = newSeq
         logger.debug("Pending sequence is now %d".format(newSeq))
         'ok
+      case SetPurgeSeqMsg(newPurgeSeq: Long) =>
+        purgeSeq = newPurgeSeq
```
rnewson: this should be `pendingPurgeSeq = newPurgeSeq`. `purgeSeq` must only change after it's been committed.

rnewson (changes requested):

src/main/scala/com/cloudant/clouseau/IndexService.scala
chguocloudant: Got it. I will modify it.

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -117,6 +121,10 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
         pendingSeq = newSeq
         logger.debug("Pending sequence is now %d".format(newSeq))
         'ok
+      case SetPurgeSeqMsg(newPurgeSeq: Long) =>
+        pendingPurgeSeq = newPurgeSeq
+        logger.debug("purge sequence is now %d".format(newPurgeSeq))
```
rnewson, line 126: should say 'Pending purge sequence is now %d'

rnewson (approved):
+1 but let's merge once the matching dreyfus PR is ready.

src/main/scala/com/cloudant/clouseau/IndexService.scala
chguocloudant: OK, got it. I will modify it later.

PR #8, opened 2016-12-01
Convert debugging [INFO] logs to more appropriate [DEBUG] logs

mayya-sharipova:


mayya-sharipova:
porting it from cloudant/clouseau, where it was already +1ed

PR #11, opened 2017-02-16
81816 clouseau for clustered purge improvements

chguocloudant:
## Overview

The primary purpose of clustered purge is to clean databases that have multiple deleted tombstones or single documents that contain large numbers of conflicts. But it can also be used to purge any document (deleted or non-deleted) with any number of revisions.

With PR https://github.com/apache/couchdb/pull/1187, the purge API and update on secondary index are implemented. The current PR is in charge of updating search index according to purge tree and record purge history in local purge document to make sure that the search index update is not out of date.

## JIRA issue number
https://issues.apache.org/jira/browse/COUCHDB-3326

https://github.com/apache/couchdb/pull/1186
https://github.com/apache/couchdb/pull/1187

## Checklist

- [X] Code is written and works correctly;
- [X] Changes are covered by tests;
- [X] Documentation reflects the changes;

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
@@ -109,6 +109,8 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
         case pid =>
           ('ok, pid)
       }
+    case ('get_root_dir) =>
+      ('ok, rootDir.getAbsolutePath())
     case ('delete, path: String) =>
       lru.get(path) match {
         case null =>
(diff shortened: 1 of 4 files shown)


src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
```diff
@@ -109,6 +109,8 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
         case pid =>
           ('ok, pid)
       }
+    case ('get_root_dir) =>
+      ('ok, rootDir.getAbsolutePath())
```
mayya-sharipova, line 113: LGTM, it would be nice if you can provide a comment where this will be used.

PR #25, opened 2019-11-21
close writer before soft deletion of searchindex

jiangphcn:
## Overview

This fix is to refactor the logic for soft-deletion of search index. In the past, there is case where write.lock was stayed in search index directory and can't be moved. It is due to the fact that the index was not closed gracefully before moving/renaming search index. With this fix, the search index writer can be closed, and then search index can be moved to .deleted directory smoothly.

## Testing recommendations

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running com.cloudant.clouseau.AnalyzerServiceSpec
Running com.cloudant.clouseau.AnalyzerServiceSpec
Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.024 sec
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.238 sec
Running com.cloudant.clouseau.ClouseauTypeFactorySpec
Running com.cloudant.clouseau.ClouseauTypeFactorySpec
Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.002 sec
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.022 sec
Running com.cloudant.clouseau.IndexManagerServiceSpec
Running com.cloudant.clouseau.IndexManagerServiceSpec
2019-11-21 15:26:30 clouseau [INFO] foo.5432109876 Renaming '/Users/jiangph/couchdb/clouseau.dreyfus/clouseau/target/indexes/foo.5432109876' to '/Users/jiangph/couchdb/clouseau.dreyfus/clouseau/target/indexes/foo.20191121.072630.deleted.5432109876'
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.445 sec
Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.465 sec
Running com.cloudant.clouseau.ClouseauQueryParserSpec
Running com.cloudant.clouseau.ClouseauQueryParserSpec
Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 sec
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.018 sec
Running com.cloudant.clouseau.IndexServiceSpec
Running com.cloudant.clouseau.IndexServiceSpec
Tests run: 23, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 23.937 sec
Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 23.954 sec
Running com.cloudant.clouseau.SupportedAnalyzersSpec
Running com.cloudant.clouseau.SupportedAnalyzersSpec
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 sec
Tests run: 42, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.109 sec

Results :

Tests run: 95, Failures: 0, Errors: 0, Skipped: 0
```

https://github.com/apache/couchdb/pull/2130

## Checklist

- [x] Code is written and works correctly
- [X] Changes are covered by tests
- [ ] Any new configurable parameters are documented in `rel/overlay/etc/default.ini`
- [ ] A PR for documentation changes has been made in https://github.com/apache/couchdb-documentation

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/IndexCleanupService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexCleanupService.scala
@@ -15,9 +15,6 @@ package com.cloudant.clouseau
 import com.yammer.metrics.scala._
 import java.io.File
 import java.util.regex.Pattern
-import java.text.SimpleDateFormat
-import java.util.Calendar
-import java.util.TimeZone
 import org.apache.log4j.Logger
 import scalang._
 
@@ -31,19 +28,10 @@ class IndexCleanupService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
       val dir = new File(rootDir, path)
       logger.info("Removing %s".format(path))
       recursivelyDelete(dir)
-    case RenamePathMsg(dbName: String) =>
-      val srcDir = new File(rootDir, dbName)
-      val sdf = new SimpleDateFormat("yyyyMMdd'.'HHmmss")
-      sdf.setTimeZone(TimeZone.getTimeZone("UTC"))
-      val sdfNow = sdf.format(Calendar.getInstance().getTime())
-      // move timestamp information in dbName to end of destination path
-      // for example, from foo.1234567890 to foo.20170912.092828.deleted.1234567890
-      val destPath = dbName.dropRight(10) + sdfNow + ".deleted." + dbName.takeRight(10)
-      val destDir = new File(rootDir, destPath)
-      logger.info("Renaming '%s' to '%s'".format(
-        srcDir.getAbsolutePath, destDir.getAbsolutePath)
-      )
-      rename(srcDir, destDir)
+    case RenamePathMsg(path: String) =>
+      logger.info("Soft-deleting " + path)
+      val pattern = Pattern.compile(path + "/([0-9a-f]+)$")
+      rename(rootDir, path, pattern)
     case CleanupDbMsg(dbName: String, activeSigs: List[String]) =>
       logger.info("Cleaning up " + dbName)
       val pattern = Pattern.compile("shards/[0-9a-f]+-[0-9a-f]+/" + dbName + "\\.[0-9]+/([0-9a-f]+)$")
@@ -78,13 +66,23 @@ class IndexCleanupService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
       fileOrDir.delete
   }
 
-  private def rename(srcDir: File, destDir: File) {
-    if (!srcDir.isDirectory) {
+  private def rename(fileOrDir: File, path: String, includePattern: Pattern) {
+    if (!fileOrDir.isDirectory) {
       return
     }
-    if (!srcDir.renameTo(destDir)) {
-      logger.error("Failed to rename directory from '%s' to '%s'".format(
-        srcDir.getAbsolutePath, destDir.getAbsolutePath))
+    for (file <- fileOrDir.listFiles) {
+      rename(file, path, includePattern)
+    }
+
+    val m = includePattern.matcher(fileOrDir.getAbsolutePath)
+    if (m.find) {
+      logger.info("Soft-deleting index " + m.group)
+      call('main, ('rename, m.group)) match {
+        case 'ok =>
+          'ok
+        case ('error, 'not_found) =>
+          Utils.rename(rootDir, path, fileOrDir.getName)
+      }
     }
   }
 
--- a/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
@@ -119,6 +119,18 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
           pid ! 'delete
           'ok
       }
+    case ('rename, path: String) =>
+      lru.get(path) match {
+        case null =>
+          ('error, 'not_found)
+        case pid: Pid =>
+          call('pid, ('close_writer))
+          val dbpath = path.substring(0, path.lastIndexOf('/'))
+          val sig = path.substring(path.lastIndexOf('/') + 1)
+          Utils.rename(rootDir, dbpath, sig)
+          call('pid, ('exit_with_deleted))
+          'ok
+      }
     case DiskSizeMsg(path: String) =>
       getDiskSize(path)
     case 'close_lru =>
--- a/src/main/scala/com/cloudant/clouseau/IndexService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexService.scala
@@ -169,6 +169,12 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
         exit("Idle Timeout")
       }
       idle = true
+    case 'close_writer =>
+      debug("Closing writer")
+      ctx.args.writer.close()
+      'ok
+    case 'exit_with_deleted =>
+      exit('deleted)
     case 'count_fields =>
       countFields
     case 'delete =>
@@ -210,7 +216,6 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
   }
 
   override def exit(msg: Any) {
-    debug("Closed with reason: %.1000s".format(msg))
     try {
       reader.close()
     } catch {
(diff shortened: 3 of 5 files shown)


src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -220,14 +227,38 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
       ctx.args.writer.rollback()
     } catch {
       case e: AlreadyClosedException => 'ignored
-      case e: IOException =>
-        warn("Error while closing writer", e)
-        ctx.args.writer.close()
+      case e: IOException => warn("Error while closing writer", e)
```
theburge: Even though we have the correct close-then-rename order, do you think it's worth including the suggested approach from the Lucene 4.6.1 docs, which was something like:

```
 try {
   writer.close();
 } finally {
   if (IndexWriter.isLocked(directory)) {
     IndexWriter.unlock(directory);
   }
 }
 ```

File: src/main/scala/com/cloudant/clouseau/IndexManagerService.scala [hunk 2]
```diff
@@ -136,6 +136,26 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
     case CloseLRUByPathMsg(path: String) =>
       lru.closeByPath(path)
       'ok
+    case SoftDeleteMsg(path: String) =>
+      lru.get(path) match {
+        case null =>
+          val manager = self
+          node.spawn((_) => {
+            openTimer.time {
+              IndexService.start(node, ctx.args.config, path, "standard") match {
```
theburge: Is this not a bit wasteful? (Imagine a bulk deletion case: it dramatically increases the intensity of the work done.)

It seems to me that here we could:

* Verify it's not in LRU
* Directly execute the rename operation on the appropriate directory in this actor

(It looks to me like `IndexService.softDelete()` is already passed almost all of the context it requires, and it wouldn't be too difficult to call it from here either.)

A downside here would be additional error handling (e.g., if the directory doesn't exist, etc.). Another would be that we might increase the time taken to execute the `'soft_delete` operations in this actor, and delay legitimate open operations, etc.

theburge (commented):
Seems reasonable to me (with some very rusty Scala knowledge). I have a few comments on the approach, but I'll defer to Bob for a final opinion.

src/main/scala/com/cloudant/clouseau/IndexService.scala
jiangphcn: Good suggestion. I can add this.

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
jiangphcn: Most of logic in `IndexService.softDelete()` can be moved to `IndexManagerService.scala`. Just one line `ctx.args.writer.close()` in https://github.com/cloudant-labs/clouseau/pull/25/files#diff-2ed27527cbacd7a17e06d19270f99182R185,  this still needs IndexService.

src/main/scala/com/cloudant/clouseau/ClouseauTypeFactory.scala
```diff
@@ -34,6 +34,7 @@ case class RenamePathMsg(dbName: String)
 case class CleanupDbMsg(dbName: String, activeSigs: List[String])
 case class DiskSizeMsg(path: String)
 case class CloseLRUByPathMsg(path: String)
+case class SoftDeleteMsg(path: String)
```
rnewson: should delete RenamePathMsg above.

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
rnewson: "soft" delete should be handled the same as "hard" delete. A message to be sent from dreyfus to the IndexService actor which, on receipt, closes the index and renames its files then calls exit on its own pid. This ensures the correct order of processing.

rnewson (changes requested):

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
rnewson: We should not duplicate part of the OpenIndexMsg logic here. I note also that this doesn't put the Index in the LRU, so could potentially lead us to attempt to open the index multiple times.

src/main/scala/com/cloudant/clouseau/IndexCleanupService.scala
```diff
@@ -31,19 +28,10 @@ class IndexCleanupService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
...
-      // for example, from foo.1234567890 to foo.20170912.092828.deleted.1234567890
-      val destPath = dbName.dropRight(10) + sdfNow + ".deleted." + dbName.takeRight(10)
-      val destDir = new File(rootDir, destPath)
-      logger.info("Renaming '%s' to '%s'".format(
-        srcDir.getAbsolutePath, destDir.getAbsolutePath)
-      )
-      rename(srcDir, destDir)
+    case SoftDeletePathMsg(path: String) =>
+      logger.info("Soft-deleting " + path)
+      val pattern = Pattern.compile(path + "/([0-9a-f]+)$")
```
rnewson, line 33: why is this a different pattern to the one we use in CleanupDbMsg?

src/main/scala/com/cloudant/clouseau/IndexCleanupService.scala
```diff
@@ -78,13 +66,24 @@ class IndexCleanupService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
...
       return
     }
-    if (!srcDir.renameTo(destDir)) {
-      logger.error("Failed to rename directory from '%s' to '%s'".format(
-        srcDir.getAbsolutePath, destDir.getAbsolutePath))
+    for (file <- fileOrDir.listFiles) {
+      softDelete(file, path, includePattern)
+    }
+
+    logger.info(fileOrDir.getAbsolutePath)
```
rnewson: remove this line.

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -220,7 +227,16 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
       ctx.args.writer.rollback()
     } catch {
       case e: AlreadyClosedException => 'ignored
-      case e: IOException => warn("Error while closing writer", e)
+      case e: IOException =>
```
rnewson, line 228: I don't think this is the right pattern. rollback() is already a close() (but without the commit), so the fallback to a failed rollback() is just the isLocked/unlock portion (the finally clause).

src/main/scala/com/cloudant/clouseau/Utils.scala
```diff
@@ -29,4 +35,33 @@ object Utils {
     new BytesRef(string)
   }
 
+  def rename(rootDir: File, dbName: String, sig: String) {
+    val logger = Logger.getLogger("clouseau.utils")
```
rnewson, line 39: the rootDir here is always the same, so don't pass it in, simply fetch it within this method. This will make it clearer which directory we are renaming.

rnewson (changes requested):

src/main/scala/com/cloudant/clouseau/IndexCleanupService.scala
jiangphcn: for `CleanupDbMsg`, the passed is pure database name while database shardname is passed for SoftDeletePathMsg, like `shards/00000000-ffffffff/foo.1234567890`. So the regexp is different for these two messages.

src/main/scala/com/cloudant/clouseau/IndexService.scala
jiangphcn: okay, let me remove close() line.

jiangphcn:
hi Bob @rnewson thanks for your comments. I addressed them. 
In addition. I mentioned that the message from https://github.com/cloudant-labs/clouseau/pull/25/commits/105b8055349c3864eebe91885c85bde4b81eba90#diff-7a2bf8eb3cb4bef2dcc50a64d66703b1R127 was not received in IndexService. I made investigation, and found that it is possible that the Pid doesn't exist in concurrent situation. In other words, when database was deleted, I can see that https://github.com/cloudant-labs/clouseau/blob/master/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala#L159 will be called. Whether message can be sent successfully from `IndexManagerService` to `IndexServer` depends on when process exits with `normal`. 

For now, I select the approach of only closing process with pid, and then move `rename` back to `IndexCleanupService`.  Would you please take a look at 3e3f686? Thank you

src/main/scala/com/cloudant/clouseau/Utils.scala
jiangphcn: It looks that there is no`ctx` in Utils.  So I keep this for now.

PR #30, opened 2020-03-31
Add ngram and edgengram support

natcohen:
It's probably a miserable attempt but at least it's an attempt! I am not sure to understand the code but this is what I came up with.

Since the ngram tokenizer takes a `Reader` as an input, I have copied the Japanese analyzer that sends `null` (for what I think is the reader) but I am not sure that it's correct. Regarding the minimum gram and maximum gram parameters, I thought I could pass it through the CouchDB index file with options `mingram` and `maxgram`.

PR #31, opened 2020-03-31
Add edgengram support

natcohen:
Attempt to add support for EdgeNGram tokenizer


rnewson:
I think if we get https://github.com/cloudant-labs/clouseau/pull/30 right then this one will follow, so will just mentor you through the other ticket for now.

natcohen:
Merged the pull request with this one to avoid conflicts: https://github.com/cloudant-labs/clouseau/pull/30

PR #39, opened 2021-04-07
feat: add simple_asciifolding analyzer

klaemo:
As discussed on the CouchDB slack I have the need for some ASCII folding, so I added new analyzer called `simple_asciifolding`.

- it is the same as the `simple` analyzer + the [ASCIIFoldingFilter](https://lucene.apache.org/core/4_6_0/analyzers-common/org/apache/lucene/analysis/miscellaneous/ASCIIFoldingFilter.html)
- the name is inspired by the [same analyzer in ElasticSearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-asciifolding-tokenfilter.html)
- I tried to add the minimum viable tests, the failure also happens on `master`

On Slack someone mentioned a more advanced filter called [ICUFoldingFilter](https://lucene.apache.org/core/4_6_0/analyzers-icu/org/apache/lucene/analysis/icu/ICUFoldingFilter.html). I don't know if it would be worth it to use that one instead?

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
+++ b/src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
@@ -12,9 +12,13 @@
 
 package com.cloudant.clouseau
 
+import java.io.Reader
 import java.util.{ Set => JSet }
 import org.apache.log4j.Logger
 import org.apache.lucene.analysis.Analyzer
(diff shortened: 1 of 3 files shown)


rnewson (approved):
neatly done.

rnewson:
if you feel like adding another analyzer for icu folding, go ahead.

PR #40, opened 2021-04-07
feat: add simple_icufolding analyzer

klaemo:
Follow up as discussed.

As far as I understood, the `ICUFoldingFilter` doesn't need an extra `LowerCaseFilter` because it already performs case folding itself.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,6 +68,11 @@
     </dependency>
     <dependency>
       <groupId>org.apache.lucene</groupId>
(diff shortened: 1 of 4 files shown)


klaemo:
Hm, not super familiar with Lucene and the Java world, but do I need to pull in https://mvnrepository.com/artifact/org.apache.lucene/lucene-analyzers-icu/4.6.1 for this? 

Should I?

PR #44, opened 2021-11-05
Setup asdf and direnv managed development environment

iilyak:
Now `cd` into the project directory would activate correct versions of build
dependencies.

Sorry for cleaning up the trailing spaces in README.md. My editor does that by default.


tonysun83:
Seems test cases are having issues connecting with clouseau during mvn test phase:

```
java.net.ConnectException: Connection refused
	at sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
	at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:741)
	at org.jboss.netty.channel.socket.nio.NioClientSocketPipelineSink$Boss.connect(NioClientSocketPipelineSink.java:401)
	at org.jboss.netty.channel.socket.nio.NioClientSocketPipelineSink$Boss.processSelectedKeys(NioClientSocketPipelineSink.java:370)
	at org.jboss.netty.channel.socket.nio.NioClientSocketPipelineSink$Boss.run(NioClientSocketPipelineSink.java:292)
	at org.jboss.netty.util.ThreadRenamingRunnable.run(ThreadRenamingRunnable.java:108)
	at org.jboss.netty.util.internal.DeadLockProofWorker$1.run(DeadLockProofWorker.java:44)
	at overlock.threadpool.ErrorLoggedThread.run(NamedThreadFactory.scala:40)
```

.envrc
```diff
@@ -0,0 +1,3 @@
+use asdf
+watch_file ".asdf"
+dotenv_if_exists .envrc.local
```
jaydoane: At one point, I had to add this get clouseau working:
```
export MAVEN_OPTS=-Dhttps.protocols=TLSv1,TLSv1.1,TLSv1.2
```
Do you not find that necessary? If so, is this the file to include it?

Makefile
```diff
@@ -0,0 +1,44 @@
...
+
+# target: tools - Bootstrap all build tools
+PHONY: tools
+tools: .asdf
+	@touch .envrc
+
+.PHONY: clouseau1
+# target: clouseau1 - Start local inistance of clouseau1 node
+clouseau1: .asdf
+	@mvn scala:run -Dlauncher=$@
```
jaydoane: Has this recently changed to require explicitly setting the Erlang distribution cookie?

jaydoane (commented):
I would like to see this land. Do you have any desire to rebase and update?

Makefile
pgj: The Erlang cookie is now optional for Clouseau.  It is the launcher interface that has changed.  It should say something like: `mvn scala:run -Dname=clouseau1`.

.envrc
iilyak: It is not needed. The maven central no longer supports TLSv1x so this workaround no longer works. 

https://central.sonatype.org/faq/tls-info/

This was needed to convince maven to use old version of TLS because old java versions didn't have support for modern TLS versions in their http client library. The zulu distribution used in this setup backported modern TLS support to java 7.48 (used here).

README.md
```diff
@@ -2,8 +2,120 @@
...
+
+## Changing tools version
+
+The list of tools are configured in a `.tool-versions` file. Make sure you
+run `make tools` every time you add new tools. You don't need to do it when
+there is just a version change.
+
 ## Configuration options
-This guide explains the various clouseau configuration options available, and how to use them to tune clouseau performance and scalability. There are two categories of clouseau options, first category is about tuning the JVM (ex: Xmx) and other category of options that go into clouseau.ini. 
+This guide explains the various clouseau configuration options available, and how to use them to tune clouseau performance and scalability. There are two categories of clouseau options, first category is about tuning the JVM (ex: Xmx) and other category of options that go into clouseau.ini.
```
iilyak, line 126: I am sorry my editor removes trailing spaces. I can "fix it" if it is an issue.

jaydoane (approved):
Tested `make` targets:
- `clean`
- `build`
- `clouseau1`

and all work as expected. Nicely done!

One small issue, during `make build` it logs this line:
```
[INFO] Dependency-reduced POM written at: /Users/jay/repos/clouseau/dependency-reduced-pom.xml
```
Should `dependency-reduced-pom.xml` be added to `.gitignore`?

iilyak:
@pgj Would it negatively impact your work on CI if I merge it?

.tool-versions
```diff
@@ -0,0 +1,3 @@
+maven 3.8.2
```
pgj, line 2: I think `direnv` is missing from here.  It refuses to work for me otherwise, at least.

Makefile
```diff
@@ -0,0 +1,290 @@
...
+  -  The asdf tool manager would need to be installed
+  -  The direnv plugin for asdf manager would need to be installed
+
+
+= Setting up asdf tool manager
+
+  You can install asdf from brew by following command.
+
+  ```
+  brew install asdf
```
pgj, line 86: Mind that it is specific to Homebrew (technically macOS), `asdf` could be installed in other ways for different environments.

Makefile
```diff
@@ -0,0 +1,290 @@
...
+- After finishing the above two steps restart your shell.
+- Run following `asdf direnv setup --shell zsh --version latest`
+- Restart your shell once again
+
+endef
+export ZSH_DARWIN_HELP
+
+
+define CONTRIBUTE_SHELL_SETUP
+  We detected $(CURRENT_SHELL) running on $(CURRENT_OS). Unfortunatelly
```
pgj: s/Unfortunatelly/Unfortunately/

Makefile
```diff
@@ -0,0 +1,290 @@
...
+clouseau3: .asdf check-epmd
+	@mvn scala:run -Dname=$@
+
+.PHONY: clean
+# target: clean - Remove build artifacts
+clean:
+	@mvn clean
+
+check-epmd:
+	@if ! lsof -i TCP:4369 | grep -q epmd ; then echo "Error: Please start 'epmd' first"; exit 1; fi
```
pgj: Is not `pgrep` ("find process by name") enough for that?  I might be wrong here, but `lsof` is usually an optional third-party software and its presence cannot be warranted.  On the contrary, `pgrep` is part of the base system.  Another option could be to not check for `empd` every time but launch it — if it is not running, it will be, otherwise that is a no-op because `empd` can detect itself and avoids running in multiple instances.

Makefile
```diff
@@ -0,0 +1,290 @@
+.PHONY: help
+# target: help - Print this help
```
pgj, line 7: That is nice, I am glad you made available it here too.

README.md
pgj: This is a good thing, I do not think we should worry about that.

README.md
```diff
@@ -2,8 +2,120 @@
...
+asdf reshim java
+```
+
+## Building distribution
+
+Just run `make` to get `target/clouseau-{version}-SNAPSHOT.zip` file.
+
+## Running locally
+
+To run single node instance use `make clouseau1`.
```
pgj, line 100: Would not it better to create a `run` target and leave the name as a parameter?  Something like this:

```console
make run NAME=clouseau1
```

It could help to remove many of the hard-wired parts.

Makefile
pgj: However, I think it could be beneficial to let the user to set the cookie still if that is needed (for example, CouchDB uses its own cookie and the default from `~/.erlang.cookie`).  This can be done by using the `with-cookie` launcher.

```shell
mvn scala:run -Dlauncher=with-cookie  -Dname=clouseau1 -Dcookie=brumbrum
```

pgj:
> @pgj Would it negatively impact your work on CI if I merge it?

No, I do not think so.

Running Clouseau in the CouchDB CI (and the respective developer setup) is supported through [the releases section of this repository](https://github.com/cloudant-labs/clouseau/releases), i.e. downloading and unpacking pre-built JARs.  It does not have to deal with the compilation.  This makes launching the Clouseau nodes faster, can work with JRE 8, depends on the JRE itself, in exchange for not being able to serve development versions from the source locally.  For that, one has to launch Clouseau as suggested in this PR for example, and separately run CouchDB configured for looking for `clouseauN`.

Makefile
iilyak: I couldn't find information on what is more portable.

- lsof
- netstat
- pgrep

It seems lsof is available on macos and most of the linux by default. I am currently checking pgrep.

iilyak: I did check lsb-release and POSIX standards. They explicitly mention that networking tools are out of the scope. So don't know what is more portable. Keep trying...

iilyak: I can chain `ps` and `grep`.

iilyak: > I can chain `ps` and `grep`

The `ps` command behaves very differently between operating systems that follow BSD and AT&T semantics.

😢

iilyak: I'll use epmd itself

```
❯ epmd -names
epmd: Cannot connect to local epmd
```

Makefile
iilyak: Cookie support is implemented in 8099855a0ab17b61df1db60b8efc75ddc75dcc8a

.tool-versions
iilyak: Added in 8987d6eb25d2913555050c2e2ea8b539561f3f0d

Makefile
iilyak: Fixed in 42827b4ee35ae4747a991264a25d5fc17b19236d

Makefile
iilyak: Addressed in e362d8792b4bd3bf3750f822c01694af94d2d8c0

README.md
iilyak: Collapsed three targets into one in 5e95fd5c9b72d0eac92bae9a519bd115ee22ce7f

iilyak: I can add generic `make run` in addition to specific 1..3 if it is useful.

pgj: That is not necessary for now.

pgj (approved):

PR #48, opened 2022-01-12
Snapshots

rnewson:
Add ability to create a snapshot of an index in a separate directory (using hard links, so must be same filesystem).

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/ExternalSnapshotDeletionPolicy.scala
+++ b/src/main/scala/com/cloudant/clouseau/ExternalSnapshotDeletionPolicy.scala
@@ -1,0 +1,70 @@
+package com.cloudant.clouseau
+
+import java.io.File
+import java.nio.file.StandardCopyOption.ATOMIC_MOVE
+import java.nio.file.Files
+import java.util.List
+import java.util.UUID
+import org.apache.lucene.index.IndexCommit
+import org.apache.lucene.index.IndexDeletionPolicy
+import org.apache.lucene.store.FSDirectory
+import scala.collection.JavaConversions._
+
+class ExternalSnapshotDeletionPolicy(dir: FSDirectory) extends IndexDeletionPolicy {
+
+  val originDir: File = dir.getDirectory
+  var lastCommit: Option[IndexCommit] = None
+
+  def snapshot(snapshotDir: File): Unit = {
+    synchronized {
+      lastCommit match {
+        case None =>
+          throw new IllegalStateException("No index commit to snapshot");
+        case Some(commit) =>
+          if (snapshotDir.exists) {
+            throw new IllegalStateException("Snapshot directory already exists")
+          }
+          /* Prepare the snapshot directory in a temporary location so we can atomically
+          rename it into place at successful completion. */
+          val tmpDir = new File(snapshotDir.getParentFile, UUID.randomUUID.toString)
+          if (!tmpDir.mkdir) {
+            throw new IllegalStateException("Failed to make temporary directory for snapshot")
+          }
+          var success = false
+          try {
+            for (filename <- commit.getFileNames) {
+              Files.createLink(new File(tmpDir, filename).toPath, new File(originDir, filename).toPath);
+            }
+            Files.move(tmpDir.toPath, snapshotDir.toPath, ATOMIC_MOVE)
+            success = true
+          } finally {
+            // Try to clean up
+            if (!success) {
+              for (file <- tmpDir.listFiles) {
+                file.delete
+              }
+              tmpDir.delete
+            }
+          }
+      }
+    }
+  }
+
+  def onInit(commits: List[_ <: IndexCommit]): Unit = {
+    keepOnlyLastCommit(commits)
+  }
+
+  def onCommit(commits: List[_ <: IndexCommit]): Unit = {
+    keepOnlyLastCommit(commits)
+  }
+
+  private def keepOnlyLastCommit(commits: List[_ <: IndexCommit]): Unit = {
+    synchronized {
+      for (commit <- commits.reverse.drop(1)) {
+        commit.delete
+      }
+      lastCommit = commits.lastOption
+    }
+  }
+
+}
--- a/src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
+++ b/src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
@@ -116,11 +116,11 @@ object SupportedAnalyzers {
       Some(new WhitespaceAnalyzer(IndexService.version))
     case "simple_asciifolding" =>
       Some(new Analyzer() {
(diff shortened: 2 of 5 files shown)


src/main/scala/com/cloudant/clouseau/ExternalSnapshotDeletionPolicy.scala
```diff
@@ -0,0 +1,70 @@
...
+import java.nio.file.StandardCopyOption.ATOMIC_MOVE
+import java.nio.file.Files
+import java.util.List
+import java.util.UUID
+import org.apache.lucene.index.IndexCommit
+import org.apache.lucene.index.IndexDeletionPolicy
+import org.apache.lucene.store.FSDirectory
+import scala.collection.JavaConversions._
+
+class ExternalSnapshotDeletionPolicy(dir: FSDirectory) extends IndexDeletionPolicy {
```
tonysun83, line 13: I'm reading up on the old blogs on how to create snapshots. I know IndexDeletionPolicy is in memory only, which doesn't help us as we want a copy of the actual disk files. Can you explain the rationale behind creating a new extended class vs using https://lucene.apache.org/core/4_6_1/core/org/apache/lucene/index/PersistentSnapshotDeletionPolicy.html?
I do notice you're creating temporary directory names so perhaps that's the reason.

src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
```diff
@@ -116,11 +116,11 @@ object SupportedAnalyzers {
       Some(new WhitespaceAnalyzer(IndexService.version))
     case "simple_asciifolding" =>
       Some(new Analyzer() {
-         def createComponents(fieldName: String, reader: Reader): TokenStreamComponents = {
-            val tokenizer: Tokenizer = new LetterTokenizer(IndexService.version, reader);
-            new TokenStreamComponents(tokenizer, new ASCIIFoldingFilter(new LowerCaseFilter(IndexService.version, tokenizer)))
-         }
-       })
+        def createComponents(fieldName: String, reader: Reader): TokenStreamComponents = {
```
tonysun83, line 119: random white space changes?

src/main/scala/com/cloudant/clouseau/ExternalSnapshotDeletionPolicy.scala
rnewson: Because PersistentSnapshotDeletionPolicy creates the snapshot inside the Lucene directory (i.e, under /srv/search_index/... for cloudant), and the list of files constituting the snapshot is written in a binary file (snapshots_X) which OS-level tools can't read.

"IndexDeletionPolicy is in memory only" is not right. The commits that are preserved in an index are durable, it's just the default policy (KeepOnlyLastCommitDeletionPolicy) that chooses to only preserve the latest one. A policy could easily be written to preserve them all.

The purpose of this new Policy is to make it easy for administrators to take a consistent backup of the index. The directory that contains the snapshot is immutable once created (hence the atomic rename to your chosen name as the last step). The contents of that directory represent the index at the time the snapshot was created, they can be copied to some external source, and, when backup is complete, the snapshot directory can be deleted, which will potentially release the files in question (they might still be part of the _current_ state of the index).

src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
rnewson: that's in a separate commit on purpose and is unrelated to snapshots, it was just annoying me as any 'mvn' run also runs the configured code formatter and some previous commit has introduced this formatting error. I can separate it into another PR if you like.

src/main/scala/com/cloudant/clouseau/ExternalSnapshotDeletionPolicy.scala
tonysun83: > "IndexDeletionPolicy is in memory only" is not right. The commits that are preserved in an index are durable, it's just the default policy (KeepOnlyLastCommitDeletionPolicy) that chooses to only preserve the latest one. A policy could easily be written to preserve them all.

My error, I meant to say SnapShotDeletionPolicy "only in memory":
https://lucene.apache.org/core/4_6_1/core/org/apache/lucene/index/SnapshotDeletionPolicy.html
"This class maintains all snapshots in-memory, and so the information is not persisted and not protected against system failures."
IIUC, if we use SnapShotDeletionPolicy then as long as the clouseau is up, lucene won't delete the commit files referenced by the snapshot. However, if clouseau crashed, then the snapshot will be lost and subsequently, the referenced commit files will eventually become stale and deleted. 
Anyways, it was for my own understanding only. Thanks for clarifying.

tonysun83: >The contents of that directory represent the index at the time the snapshot was created, they can be copied to some external source, and, when backup is complete, the snapshot directory can be deleted, which will potentially release the files in question (they might still be part of the current state of the index).

Should we add a function to remove the snapshot/delete the files? Right now,  an operator would have to manually go in and delete the files (or write their own script to clean it up). It seems having a function would allow easier automation in the future so there aren't lingering snapshots around. I'm envisioning a scenario where snapshots are copied into cold storage but we forget to delete the snapshots on the nodes.

tonysun83:
Just had one more question regarding snapshot removal (see above thread).

Out of scope for this PR, but I'm curious whether this new functionality will be consumed with Dreyfus. From a couchdb perspective, could these snapshots also help with large indexing rebuilds? In other words, instead of having to rebuild everything for whatever reason, we can just start at the snapshot?

src/main/scala/com/cloudant/clouseau/ExternalSnapshotDeletionPolicy.scala
rnewson: aha, ok. yes, correct, SnapshotDeletionPolicy just maintains the list of IndexCommit's to preserve in memory, and all snapshots would be released on JVM exit. This is often fine, of course, and the benefit there is allowing a client to run queries over the same version of the index for a period of time, even if the index has subsequently changed.

rnewson: For cleanup, it is as simple as deleting the files in the directory specified in the {create_snapshot, Directory} location so can be left to the administrator.

rnewson:
Hi, Yes indeed. The purpose here is to allow administrators the ability to speed up database recovery. While restoring copies of the .couch files is sufficient to get you back and running it does mean waiting for all the indexes. The snapshots here will include the embedded CommitData which includes the update sequence we've reached, so clouseau will just index from that point and not zero.

As for dreyfus, I'm not sure if the api goes there as the motivation here is to add this ability for IBM/Cloudant usage in the first instance. Putting that in dreyfus means committing to a method that might not suit everyone and might change in an incompatible way in the near future.

tonysun83:
+1

PR #49, opened 2022-02-09
Make it possible to snapshot more easily

rnewson:
We can now snapshot by talking to clouseau directly.

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
@@ -13,6 +13,7 @@
 package com.cloudant.clouseau
 
 import java.io.File
+import java.io.IOException
 import java.util.HashMap
 import java.util.LinkedHashMap
 import java.util.{ Map => JMap }
@@ -139,6 +140,13 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
       'ok
     case 'version =>
       ('ok, getClass.getPackage.getImplementationVersion)
+    case ('create_snapshot, indexName: String, snapshotDir: String) =>
+      lru.get(indexName) match {
+        case null =>
+          createSnapshot(indexName, snapshotDir)
+        case pid: Pid =>
+          call(pid, ('create_snapshot, snapshotDir))
+      }
   }
 
   override def handleInfo(msg: Any) = msg match {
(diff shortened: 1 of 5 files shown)


src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
```diff
@@ -139,6 +140,13 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
       'ok
     case 'version =>
       ('ok, getClass.getPackage.getImplementationVersion)
+    case ('create_snapshot, originDir: String, snapshotDir: String) =>
```
tonysun83: I didn't consider this in the first review, but I'm curious what performance issues we might encounter by making this a synchronous call. 

From this old blog: https://freecontent.manning.com/hot-backups-with-lucene/, it  seems they call snapshot in a different thread:

```
You can do this from a separate thread, and continue using the writer as usual in your application to make changes to the index. The backup will copy the point-in-time snapshot as of the moment when you called the snapshot() method.
```

The way I'm reading the code now, either the IndexManagerService or the IndexService process serves the request. If we assume a common scenario where disk space is slow, with lots of different searches running (meaning lots of LRU evictions and opening of new indexes), then running snapshots bottle up the processes even more. 

If we made the call asynchronous though, then we wouldn't know if the snapshot succeeded from the callers perspective. Maybe have the snapshot call be asychronous and another synchronous call to check the status of the snapshot?

rnewson: If we make it asynchronous we invite the problem that the index is opened during the snapshotting process potentially yielding a set of files that do not constitute a valid index. Doing it in the IndexManagerService loop guarantees the index is either open or closed for the duration. The snapshotting process itself should be very fast. It is just creating hard links to a handful of files, each of which should take almost no time at all.

It may prove to be a problem in practice but I'd rather see it, in performance testing, first.

tonysun83:
I re-ran the java ci checks a few times and IndexService test keeps failing. It doesn't seem you modified anything that different from the previous commit.

rnewson:
hm, annoying. test suite passes locally and (clearly) the changes don't touch anything to do with highlighting. It must be a timeout somewhere. I'll keep retrying for a bit.

rnewson:
@tonysun83 fifth time lucky. I merged just to move forward. If there's problems with this in testing I'll have another PR and/or if you have review comments.

PR #59, opened 2022-11-01
don't consider an index idle if it has pending merges

rnewson:


rnewson:
This is intended to ensure Clouseau allows segment merges to happen (unless evicted by LRU) so that indexes don't accumulate large numbers of segments, which degrades performance initially and then affects service availability.

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -168,7 +168,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
     case ('close, reason) =>
       exit(reason)
     case ('close_if_idle) =>
-      if (idle) {
+      if (idle && !ctx.args.writer.hasPendingMerges()) {
```
iilyak, line 171: Just wanted to note that `hasPendingMerges` is an *experimental* API. Nevertheless it was present in 4.6.1 and is still present in 8.7.1. 

I think it is a safe bet to rely on it being stable enough.

iilyak: I have a stylistic question. Do we need `()` here? I've noticed that other functions in the same module do not have them. For example calls to [`getDirectory`](https://github.com/cloudant-labs/clouseau/pull/59/files#diff-9693dd93652a8da5c752401f5a2abd8c87750a85bbb9dfc661c700151cfcd6e9R178)

I don't have strong preference either way, but was curious. So I started to dig. Here in [stackoverflow](https://stackoverflow.com/questions/1181533/what-are-the-precise-rules-for-when-you-can-omit-parenthesis-dots-braces-f), someone is giving a quote from a book.

> By convention, functions with no arguments that have side effects, such as println, are called with parentheses; those without side effects are called without parentheses."

I don't recall any convention guide for clouseau, so it is a +1 regardless.

iilyak (approved):
+1

src/main/scala/com/cloudant/clouseau/IndexService.scala
rnewson: err, probably not, no. feel free to PR a change if it bothers you.

PR #65, opened 2023-01-19
addressing CVE-2014-3488

danydrouin:
https://github.com/advisories/GHSA-9959-6p3m-wxpc
https://nvd.nist.gov/vuln/detail/CVE-2014-3488


rnewson (approved):
As clouseau doesn't use ssl this seems fine, and we can't easily upgrade netty to the fixed version.

PR #67, opened 2023-02-08
Make Erlang cookie parametric

pgj:
When linked to CouchDB nodes, the Erlang cookie might be different from the one that is originally assumed.  Make it work with `~/.erlang.cookie` on launching the instance to facilitate smoother integration.

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/Main.scala
+++ b/src/main/scala/com/cloudant/clouseau/Main.scala
@@ -42,7 +42,7 @@ object Main extends App {
   config.addConfiguration(reloadableConfig)
 
   val name = config.getString("clouseau.name", "clouseau@127.0.0.1")
-  val cookie = config.getString("clouseau.cookie", "monster")
+  val cookie = config.getString("clouseau.cookie")
   val closeIfIdleEnabled = config.getBoolean("clouseau.close_if_idle", false)
   val idleTimeout = config.getInt("clouseau.idle_check_interval_secs", 300)
   if (closeIfIdleEnabled) {
@@ -52,7 +52,7 @@ object Main extends App {
     typeFactory = ClouseauTypeFactory,
     typeEncoder = ClouseauTypeEncoder,
     typeDecoder = ClouseauTypeDecoder)
-  val node = Node(name, cookie, nodeconfig)
+  val node = if (cookie != null) Node(name, cookie, nodeconfig) else Node(name, nodeconfig)
 
   ClouseauSupervisor.start(node, config)
   logger.info("Clouseau running as " + name)
--- a/pom.xml
+++ b/pom.xml
@@ -156,30 +156,20 @@
         <configuration>
           <launchers>
             <launcher>
-              <id>clouseau1</id>
+              <id>standard</id>
               <mainClass>com.cloudant.clouseau.Main</mainClass>
               <jvmArgs>
-                <jvmArg>-Dclouseau.name=clouseau1@127.0.0.1</jvmArg>
-                <jvmArg>-Dclouseau.cookie=monster</jvmArg>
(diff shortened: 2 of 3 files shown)


rnewson:
history: in couchdb we recently prohibited the value `monster` for security reasons (many users left this value in place and their installs were exploited as a consequence). We did not follow through with how the dev launch for clouseau works.

Like dev/run in couchdb I suggest we follow this strategy;

1) we default to not explicitly setting the cookie, and let clouseau read the ~/.erlang.cookie file instead (same as dev/run)
2) Optionally you can specify the cookie, in which case the user is responsible for setting the same value for couchdb as for clouseau.

Implied in this is that the word `monster` will no longer appear in pom.xml.

pgj:
How can `clouseau` read the `~/.erlang.cookie` file?  Does this happen when the `clouseau.cookie` variable is not set?

pgj:
Hrm, [this does not seem to be the case](https://github.com/cloudant-labs/clouseau/blob/7a601e96c549a73abce72c640bee8cd921167bcf/src/main/scala/com/cloudant/clouseau/Main.scala#L45).  Do you think I should the change Scala code to get the default value for the `cookie` from `~/.erlang.cookie`?

rnewson:
I'm thinking of https://github.com/cloudant/scalang/blob/894bd33fc563b54c455fd78ae6ca72a2d2d25811/src/main/scala/scalang/Node.scala#L68

pgj:
> I'm thinking of https://github.com/cloudant/scalang/blob/894bd33fc563b54c455fd78ae6ca72a2d2d25811/src/main/scala/scalang/Node.scala#L68

Looks like the referenced functionality only kicks in when the `cookie` parameter is not passed along with the `NodeConfig` but that latter is Clouseau-specific and cannot be omitted.

pgj:
@rnewson do you have any comments or requests on the proposed changes?  How do you like them?

src/main/scala/com/cloudant/clouseau/Main.scala
```diff
@@ -55,5 +55,6 @@ object Main extends App {
   val node = Node(name, cookie, nodeconfig)
 
   ClouseauSupervisor.start(node, config)
+  logger.info("Erlang cookie in use: " + cookie)
```
rnewson: the cookie value is a shared secret, it's not appropriate to print it to screen/log files.

rnewson (changes requested):
I don't think we should re-implement the code already presents in scalang. I'd modify Node.scala in scalang to add the missing `def apply(name : String, nodeConfig : NodeConfig)` and have it use `findOrGenerateCookie`, then construct the `Node` in clouseau's `Main` with those arguments if `clouseau.cookie` is not set.

pgj:
When I was looking at the `scalang` repository I had the impression that it is not actively maintained (the latest commit is from 9 years ago) that is why I decided not to hack on it in the first place.  But in general, I think you are right @rnewson -- I can give them a try because your way is closer to the true path.

pgj:
Opened https://github.com/boundary/scalang/pull/37 to address the `scalang` part.

jaydoane:
> When I was looking at the `scalang` repository I had the impression that it is not actively maintained (the latest commit is from 9 years ago)

We're using [this fork of scalang](https://github.com/cloudant/scalang/) which has more recent activity.

pgj:
Thanks @jaydoane for the excellent hint -- I should have thought about it.  Per this suggestion, moved the pull request to https://github.com/cloudant/scalang/pull/3 .

pgj:
@rnewson rebased the solution to `scalang 1.0.1`.  Hopefully this version is now closer to the one you suggested.

rnewson (approved):
nice.

PR #72, opened 2023-05-12
improve highlighting support

rnewson:
* gracefully handle highlighting attempts on non-stored fields
* highlight duplicate fields (need to check if this breaks compatibility before merging)

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/IndexService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexService.scala
@@ -734,9 +734,10 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
     if (HPs != null) {
       val highlights = (HPs.highlightFields zip HPs.analyzers).map {
         case (field, analyzer) =>
-          val frags = HPs.highlighter.getBestFragments(analyzer, field,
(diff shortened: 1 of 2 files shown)


rnewson:
Think we're safe compatibility wise. It's just a longer list.

rnewson:
reporter confirms fix works as expected. (cc @jaydoane )

jaydoane (approved):
+1 tests pass and LGTM

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -734,9 +734,10 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
     if (HPs != null) {
       val highlights = (HPs.highlightFields zip HPs.analyzers).map {
         case (field, analyzer) =>
-          val frags = HPs.highlighter.getBestFragments(analyzer, field,
-            doc.get(field), HPs.highlightNumber).toList
-          (field, frags)
+          (field, doc.getValues(field).flatMap { v =>
```
pgj, line 737: I like better the original approach of using the additional `frags` variable -- in my humble opinion this makes the code easier to read.

PR #74, opened 2023-05-20
feat: uplift Lucene to 4.10.4

pgj:
Update Clouseau to use the latest available Lucene version from the 4.x series.  This change brings many bug fixes while it maintains the backward compatibility for the existing indexes.  4.10.4 has 45 API changes and 116 bug fixes since 4.6.1, for the details see the [change log](https://lucene.apache.org/core/4_10_4/changes/Changes.html).

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/ClouseauTypeFactory.scala
+++ b/src/main/scala/com/cloudant/clouseau/ClouseauTypeFactory.scala
@@ -17,15 +17,15 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.apache.lucene.document.Field._
 import org.apache.lucene.document._
+import org.apache.lucene.facet.FacetsConfig
+import org.apache.lucene.facet.sortedset.SortedSetDocValuesFacetField
+import org.apache.lucene.index.FieldInfo.IndexOptions
 import org.apache.lucene.search._
+import org.apache.lucene.util.BytesRef
 import scala.collection.immutable.Map
 import scala.collection.JavaConversions._
 import scalang._
 import org.jboss.netty.buffer.ChannelBuffer
-import org.apache.lucene.util.BytesRef
-import org.apache.lucene.facet.params.FacetIndexingParams
-import org.apache.lucene.facet.sortedset.SortedSetDocValuesFacetFields
-import org.apache.lucene.facet.taxonomy.CategoryPath
 import scala.collection.mutable.ArrayBuffer
 
 case class SearchRequest(options: Map[Symbol, Any])
@@ -51,6 +51,22 @@ case class SetPurgeSeqMsg(seq: Long)
 object ClouseauTypeFactory extends TypeFactory {
 
   val logger = LoggerFactory.getLogger("clouseau.tf")
+
+  // Default numeric precision step changed in Lucene 4.9, this
+  // wrapper is to help to preserve the original setting for
+  // compatibility reasons :-(
+  def maybeDoubleStored(stored: Store) = {
+    val result = new FieldType()
+    result.setIndexed(true)
+    result.setTokenized(true)
+    result.setOmitNorms(true)
+    result.setIndexOptions(IndexOptions.DOCS_ONLY)
+    result.setNumericType(FieldType.NumericType.DOUBLE)
+    result.setNumericPrecisionStep(8)
+    result.setStored(stored == Store.YES)
+    result.freeze()
+    result
+  }
 
   def createType(name: Symbol, arity: Int, reader: TermReader): Option[Any] = (name, arity) match {
     case ('open, 4) =>
@@ -113,7 +129,8 @@ object ClouseauTypeFactory extends TypeFactory {
     for (field <- reader.readAs[List[Any]]) {
       addFields(result, field)
     }
-    result
+    val fc = new FacetsConfig()
+    fc.build(result)
   }
 
   private def addFields(doc: Document, field0: Any): Unit = field0 match {
@@ -129,12 +146,7 @@ object ClouseauTypeFactory extends TypeFactory {
           }
           doc.add(field)
           if (isFacet(map) && value.nonEmpty) {
-            val fp = FacetIndexingParams.DEFAULT
-            val delim = fp.getFacetDelimChar
-            if (!name.contains(delim) && !value.contains(delim)) {
-              val facets = new SortedSetDocValuesFacetFields(fp)
-              facets.addFields(doc, List(new CategoryPath(name, value)))
-            }
+            doc.add(new SortedSetDocValuesFacetField(name, value))
           }
         case None =>
           'ok
@@ -151,7 +163,7 @@ object ClouseauTypeFactory extends TypeFactory {
       val map = options.toMap
       toDouble(value) match {
         case Some(doubleValue) =>
-          doc.add(new DoubleField(name, doubleValue, toStore(map)))
+          doc.add(new DoubleField(name, doubleValue, maybeDoubleStored(toStore(map))))
           if (isFacet(map)) {
             doc.add(new DoubleDocValuesField(name, doubleValue))
           }
--- a/src/main/scala/com/cloudant/clouseau/IndexService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexService.scala
@@ -26,6 +26,7 @@ import org.apache.lucene.util.BytesRef
 import org.apache.lucene.util.Version
 import org.apache.lucene.search.IndexSearcher
 import org.apache.lucene.search.BooleanClause.Occur
+import org.apache.lucene.search.MultiCollector
 import org.apache.lucene.queryparser.classic.QueryParser
 import org.apache.lucene.queryparser.classic.ParseException
 import org.apache.lucene.search.highlight.{
@@ -40,18 +41,11 @@ import collection.JavaConversions._
 import com.yammer.metrics.scala._
 import com.cloudant.clouseau.Utils._
 import org.apache.commons.configuration.Configuration
-import org.apache.lucene.facet.sortedset.{
-  SortedSetDocValuesReaderState,
-  SortedSetDocValuesAccumulator
-}
-import org.apache.lucene.facet.range.{
-  DoubleRange,
-  RangeAccumulator,
-  RangeFacetRequest
-}
-import org.apache.lucene.facet.search._
+import org.apache.lucene.facet.DrillDownQuery
+import org.apache.lucene.facet.{ Facets, FacetsCollector, FacetsConfig, FacetResult }
+import org.apache.lucene.facet.sortedset.{ DefaultSortedSetDocValuesReaderState, SortedSetDocValuesFacetCounts }
+import org.apache.lucene.facet.range.{ DoubleRange, DoubleRangeFacetCounts }
 import org.apache.lucene.facet.taxonomy.CategoryPath
-import org.apache.lucene.facet.params.{ FacetIndexingParams, FacetSearchParams }
 import scala.Some
 import scalang.Pid
 import com.spatial4j.core.context.SpatialContext
@@ -302,27 +296,19 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
               baseQuery
             case categories: List[List[String]] =>
               val drilldownQuery = new DrillDownQuery(
-                FacetIndexingParams.DEFAULT, baseQuery)
+                new FacetsConfig, baseQuery)
               for (category <- categories) {
-                val category1 = category.toArray
-                val len = category1.length
-                try {
-                  if (len < 3) {
-                    drilldownQuery.add(new CategoryPath(category1: _*))
-                  } else { //if there are multiple values OR'd them, delete this else part after updating to Apache Lucene > 4.6
-                    val dim = category1(0)
-                    val categoryPaths: Array[CategoryPath] = new Array[CategoryPath](len - 1)
-                    for (i <- 1 until len) {
-                      categoryPaths(i - 1) = new CategoryPath(Array(dim, category1(i)): _*)
-                    }
-                    drilldownQuery.add(categoryPaths: _*)
+                val dim = category.head
+                for (path <- category.tail) {
+                  try {
+                    drilldownQuery.add(dim, path)
+                  } catch {
+                    case e: IllegalArgumentException =>
+                      throw new ParseException(e.getMessage)
+                    case e: ArrayStoreException =>
+                      throw new ParseException(category +
+                        " contains a non-string item")
                   }
-                } catch {
-                  case e: IllegalArgumentException =>
-                    throw new ParseException(e.getMessage)
-                  case e: ArrayStoreException =>
-                    throw new ParseException(category +
-                      " contains a non-string item")
                 }
               }
               drilldownQuery
@@ -352,39 +338,9 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
                 false, docsScoredInOrder)
           }
 
-          val countsCollector = createCountsCollector(counts)
-
-          val rangesCollector = ranges match {
-            case None =>
-              null
-            case Some(rangeList: List[_]) =>
-              val rangeFacetRequests = for ((name: String, ranges: List[_]) <- rangeList) yield {
-                new RangeFacetRequest(name, ranges.map({
-                  case (label: String, rangeQuery: String) =>
-                    ctx.args.queryParser.parse(rangeQuery) match {
-                      case q: NumericRangeQuery[_] =>
-                        new DoubleRange(
-                          label,
-                          ClouseauTypeFactory.toDouble(q.getMin).get,
-                          q.includesMin,
-                          ClouseauTypeFactory.toDouble(q.getMax).get,
-                          q.includesMax)
-                      case _ =>
-                        throw new ParseException(rangeQuery +
-                          " was not a well-formed range specification")
-                    }
-                  case _ =>
-                    throw new ParseException("invalid ranges query")
-                }))
-              }
-              val acc = new RangeAccumulator(rangeFacetRequests)
-              FacetsCollector.create(acc)
-            case Some(other) =>
-              throw new ParseException(other + " is not a valid ranges query")
-          }
-
+          val facetsCollector = new FacetsCollector()
           val collector = MultiCollector.wrap(
-            hitsCollector, countsCollector, rangesCollector)
+            facetsCollector, hitsCollector)
 
           searchTimer.time {
             partition match {
@@ -412,12 +368,14 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
           if (legacy) {
             ('ok, TopDocs(updateSeq, getTotalHits(hitsCollector), hits))
           } else {
+            val countsFacets = pullCountsFacets(counts, facetsCollector)
+            val rangesFacets = pullRangesFacets(ranges, facetsCollector)
             ('ok, List(
               ('update_seq, updateSeq),
               ('total_hits, getTotalHits(hitsCollector)),
               ('hits, hits)
-            ) ++ convertFacets('counts, countsCollector)
-              ++ convertFacets('ranges, rangesCollector))
+            ) ++ convertFacets('counts, countsFacets)
+              ++ convertFacets('ranges, rangesFacets))
           }
         }
       case error =>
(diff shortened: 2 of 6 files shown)


pgj:
It passes the `make elixir-search` tests, but many tests fail when running with `make mango-test`:

```console
COUCH_USER=adm COUCH_PASS=pass .venv/bin/python3 -m nose2  ... ........s.ss.s.s.............s.ss.s.s.......F.FF.s.ss.s.s........................................................................................................................................FF...................s...................................................FF...............................................................................F.....FF.......
======================================================================
FAIL: test_elem_match (03-operator-test.OperatorTextTests.test_elem_match)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/03-operator-test.py", line 48, in test_elem_match
    self.assertEqual(len(docs), 1)
AssertionError: 0 != 1

======================================================================
FAIL: test_eq_null_does_not_include_missing (03-operator-test.OperatorTextTests.test_eq_null_does_not_include_missing)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/03-operator-test.py", line 113, in test_eq_null_does_not_include_missing
    self.assertUserIds(user_ids, docs)
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/03-operator-test.py", line 22, in assertUserIds
    self.assertEqual(user_ids, user_ids_returned)
AssertionError: Lists differ: [9] != []

First list contains 1 additional elements.
First extra element 0:
9

- [9]
?  -

+ []

======================================================================
FAIL: test_exists_false (03-operator-test.OperatorTextTests.test_exists_false)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/03-operator-test.py", line 106, in test_exists_false
    self.assertUserIds(user_ids, docs)
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/03-operator-test.py", line 22, in assertUserIds
    self.assertEqual(user_ids, user_ids_returned)
AssertionError: Lists differ: [2, 3, 5, 6, 7, 8, 10, 11, 12, 14] != []

First list contains 10 additional elements.
First extra element 0:
2

- [2, 3, 5, 6, 7, 8, 10, 11, 12, 14]
+ []

======================================================================
FAIL: test_gt (06-basic-text-test.BasicTextTests.test_gt)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/06-basic-text-test.py", line 152, in test_gt
    assert len(docs) == 2
           ^^^^^^^^^^^^^^
AssertionError

======================================================================
FAIL: test_gte (06-basic-text-test.BasicTextTests.test_gte)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/06-basic-text-test.py", line 175, in test_gte
    assert len(docs) == 2
           ^^^^^^^^^^^^^^
AssertionError

======================================================================
FAIL: test_limit_bookmark (08-text-limit-test.LimitTests.test_limit_bookmark)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/08-text-limit-test.py", line 97, in test_limit_bookmark
    self.run_bookmark_check(i)
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/08-text-limit-test.py", line 115, in run_bookmark_check
    assert len(seen_docs) == len(limit_docs.DOCS)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError

======================================================================
FAIL: test_limit_field (08-text-limit-test.LimitTests.test_limit_field)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/08-text-limit-test.py", line 23, in test_limit_field
    assert len(docs) == 8
           ^^^^^^^^^^^^^^
AssertionError

======================================================================
FAIL: test_guess_dup_type_sort (09-text-sort-test.SortTests.test_guess_dup_type_sort)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/09-text-sort-test.py", line 79, in test_guess_dup_type_sort
    self.assertEqual(len(docs), 15)
AssertionError: 0 != 15

======================================================================
FAIL: test_number_sort (09-text-sort-test.SortTests.test_number_sort)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/09-text-sort-test.py", line 22, in test_number_sort
    self.assertEqual(len(docs), 15)
AssertionError: 0 != 15

======================================================================
FAIL: test_number_sort_desc (09-text-sort-test.SortTests.test_number_sort_desc)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/gaborpali/projects/github/couchdb/src/mango/test/09-text-sort-test.py", line 28, in test_number_sort_desc
    self.assertEqual(len(docs), 15)
AssertionError: 0 != 15

----------------------------------------------------------------------
Ran 362 tests in 26.674s

FAILED (failures=10, skipped=16)
```

The common reason for the failures is that Clouseau does not return any value.  Based on my investigations so far, it works as expected up to 4.8.1, and it starts to fail for 4.9.0 and later versions.  I have checked the change log, but none of the API changes directly indicate anything related.

jaydoane:
Awesome work!

I tested it, and see the same test failures:
```
Ran 361 tests in 39.905s

FAILED (failures=10, skipped=16)
```
Also noting that my machine appears to be somewhat slower than yours (my ~40s vs your ~27s), which I wonder if accounts for why the write lock timeout error seems more reliable on your hardware?

FWIW, I also ran these 2 tests 1000 times in a row, and didn't see any errors:
```
test_disable_index_array_length (10-disable-array-length-field-test.DisableIndexArrayLengthsTest) ... ok
test_enable_index_array_length (10-disable-array-length-field-test.DisableIndexArrayLengthsTest) ... ok
```

pgj:
> Awesome work!

👍   But, as indicated, I will have to nag @rnewson for hints on why it does not fully work.

> I tested it, and see the same test failures:

Thanks for the confirmation!

> [M]y machine appears to be somewhat slower than yours [..], which I wonder if accounts for why the write lock timeout error seems more reliable on your hardware?

Yes, that is possible.  However, I thought that you have an 2021 M1 Max too.

> FWIW, I also ran these 2 tests 1000 times in a row, and didn't see any errors:

You will not, these tests are faulty.  And I think I have found the cause for that -- I will soon submit a PR about that for `apache/couchdb`.  Prepare for the facepalm :-D

jaydoane:
> these tests are faulty

Yeah, I've been debugging them a bit today. Seems like the user docs get bulk loaded into the db, but then db info only shows the ddocs, but also no deleted docs. I'm looking forward to the 🤦

pgj:
Here are the bugs of `10-disable-array-length-field-test` with their proposed fixes: https://github.com/apache/couchdb/pull/4610.

pgj:
One of the queries that return 0 hits starting from 4.9.0:

`03-operator-test.test_elem_match`, Mango query:

```json
{
  "_id": {"$gt": null},
  "bang": {"$elemMatch": {"foo": {"$gte": 1}, "bam": true}},
}
```

which translates to:

```
($fieldnames:_5fid_3a* $fieldnames:_5fid.*) +(+bang._5b_5d.foo_3anumber:[1.0 TO Infinity] +bang._5b_5d.bam_3aboolean:true)
```

Note that the following queries work at the same time:

```
($fieldnames:_5fid_3a* $fieldnames:_5fid.*) +bang._5b_5d.foo_3anumber:[1.0 TO Infinity]
```

and

```
($fieldnames:_5fid_3a* $fieldnames:_5fid.*) +bang._5b_5d.bam_3aboolean:true
```

pgj:
After taking a closer look at the rest of the failing tests cases, I noticed that it is the ranges on numbers that cause the problems.  More specifically, ranges with no lower or upper limits, i.e. the use of the `Infinity` constant.  Apparently, Lucene 4.9 and higher do not handle properly if `Infinity` is written with a capital `I`.

The previous query works if the `Infinity` is either written as `infinity` or `infinite`, that is:

```
($fieldnames:_5fid_3a* $fieldnames:_5fid.*) +(+bang._5b_5d.foo_3anumber:[1.0 TO infinity] +bang._5b_5d.bam_3aboolean:true)
```

If CouchDB is modified according to this observation, `make mango-test` passes:

```diff
diff --git a/src/mango/src/mango_selector_text.erl b/src/mango/src/mango_selector_text.erl
index 1f8609ac2..4b46031eb 100644
--- a/src/mango/src/mango_selector_text.erl
+++ b/src/mango/src/mango_selector_text.erl
@@ -304,11 +304,11 @@ range(gt, Arg) ->
     [<<"{">>, value_str(Arg), <<" TO ", Max/binary, "]">>].

 get_range(min, Arg) when is_number(Arg) ->
-    <<"-Infinity">>;
+    <<"-infinity">>;
 get_range(min, _Arg) ->
     <<"\"\"">>;
 get_range(max, Arg) when is_number(Arg) ->
-    <<"Infinity">>;
+    <<"infinity">>;
 get_range(max, _Arg) ->
     <<"\u0x10FFFF">>.
```

pgj:
Unfortunately, this latest fix does not solve the problem, only masks it.  That is because using `infinity` (basically any other text literal) causes Lucene to return all the documents which are then filtered on the CouchDB side by Mango.  That is, more documents are returned than required in result and this is not efficient.

I was playing further with the Lucene queries.  For the `user_docs` sample database used by the Mango integration tests, I noticed the following behavior.  Documents are there and all of them are returned when no filtering is applied:

```
*:*
```

`Infinity` as upper bound works when the lower bound is set to the minimum `age` value, e.g.:

```
age_3anumber:[22.0 TO Infinity]
```

This can be lowered until `2.0` when Lucene stops returning any hit, e.g.:

```
age_3anumber:[2.0 TO Infinity] 
```

When `Infinity` is not used, even `0.0` (as used in many tests) works as lower bound, e.g.:

```
age_3anumber:[0.0 TO 100.0]
```

At the same time, the inclusive ranges do not include the value of the lower bound.  That is the following query will not include `73` unless the lower bound is lowered to `72.0`:

```
age_3anumber:[73.0 TO Infinity]
```

pgj:
For the reference, here are the commands that I used for the investigation:

```shell
# Run in the CouchDB git clone:
MANGO_TESTS_KEEP_DBS=1 make mango-test MANGO_TEST_OPTS="03-operator-test.OperatorTextTests.test_elem_match"
# Start CouchDB manually via dev/run (-a adm:pass -n 1), then:
MANGO_DB=$(curl -sS http://adm:pass@localhost:15984/_all_dbs | jq -r '.[]|select(startswith("mango_test_"))')
# Set the query:
QUERY="age_3anumber:[2.1 TO Infinity]"
# Run the query:
curl -s -X POST -H 'Content-Type: application/json' "http://adm:pass@localhost:15984/$MANGO_DB/_design/cedc01a027213706d7260b5e5b73c70b9233743a/_search/cedc01a027213706d7260b5e5b73c70b9233743a" -d '{"q":"'$QUERY'","include_docs":true}' | jq -r '.'
```

pgj:
I think I have the source of the problem.  [LUCENE-5609](https://issues.apache.org/jira/browse/LUCENE-5609) applies an optimization which technically implies a change in the run-time behavior because `DoubleRange`s should be created with the precision step of 16 instead of 8.  I changed it to this value but we may want to lose it completely, because it is optional.

src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
```diff
@@ -62,7 +62,7 @@ class ClouseauQueryParser(version: Version,
                              startInclusive: Boolean,
                              endInclusive: Boolean): Query = {
     if (isNumber(lower) && isNumber(upper)) {
-      NumericRangeQuery.newDoubleRange(field, 8, lower.toDouble,
+      NumericRangeQuery.newDoubleRange(field, 16, lower.toDouble,
```
rnewson: This triggered some investigation on my part, coupled with the other comments about precision step default change.

Seems we currently index with precision step of 4 (default for 4.6.1) but query with precision step 8. That's probably a bug, albeit one we've had for a long time.

I don't think we can change this when we upgrade as it might cause search results to change.

Instead, we need to keep the above as `8` and change how we _index_ doubles to continue to do so with precision step of 4 not the 'new' default of 16.

rnewson: See https://github.com/cloudant-labs/clouseau/commit/742670bc9395ac458edbcbc1fdc8639610634dcb for my attempt at doing so.

pgj: I have had a feeling that this is going to be something that is problematic.  The optimization leads to smaller indexes which means they need to be rebuilt, right?  The integration tests work with fresh indexes, but for existing deployments this could be a problem.

Thank you very much for the proposal, I will have take a look at it and integrate soon.

rnewson: https://github.com/cloudant-labs/clouseau/commit/3519e6803f97ad8596b26bf72bb40472acef3b2a rather

rnewson: no, it means that search results could change. some double values will be quantised to different values before and after this change. Existing indexes will have a mixture of the old and new precision step. Simply put, we can change precision step (and other things) but it requires an index rebuild to get the right answers.

rnewson: I looked into it further, I think we would benefit from indexing with precision step `8`, given that we have (mistakenly) been using that value in all numeric range queries. Creating extra terms with precision step `4` is wasteful as we never use them for searching.

```
tokens created for precision step 4;
0 4653142547559993678
4 4653142547559993664
8 4653142547559993600
12 4653142547559993344
16 4653142547559940096
20 4653142547559874560
24 4653142547550437376
28 4653142547349110784
32 4653142546006933504
36 4653142485877391360
40 4653142004841054208
44 4653133208748032000
48 4653062840003854336
52 4652218415073722368
56 4611686018427387904
60 4611686018427387904

tokens created for precision step 8;
0 4653142547559993678
8 4653142547559993600
16 4653142547559940096
24 4653142547550437376
32 4653142546006933504
40 4653142004841054208
48 4653062840003854336
56 4611686018427387904
```

rnewson: also quoting from 5609;

> To be simple, i think a change we make has to be a multiple of 4: the current default.
>This way we have easy backwards compat with existing indexes for users that just go with the default (since NumericRangeQuery has ctors that use the default precisionStep, and multiples of the index-time work at query-time).

pgj: Sorry, I do not see how does all this translates to the current situation.  Should we change the `Double` field type wrapper to set the numeric precision step to `8` instead of `4`?

rnewson: Sorry, there are several issues here.

Clouseau has apparently always been wrong on this. We are indexing numeric values with precision step of 4 but only ever querying with precision step of 8. Half of those terms are never queried, there's no possible path to them, but we took the time to index them and store them.

Later version of Lucene have changed the default (from 4 to 16), which triggered this conversation.

I don't think we should change to 16 as it would change the results of current queries (I think, but this should be challenged, perhaps I misunderstand the implications).

So, the least we must do is prevent the default from changing to 16, and that requires the FieldType changes I show in my commit above.

The only question remaining is whether we keep exactly what we have today (4 for indexing, 8 for querying), or whether we address that mistake and switch to 8 for indexing and querying.

If I'm wrong about the implications of 16 then we could of course do nothing, but someone other than me will need to do that investigation, or it's just an echo chamber.

Specifically, I worry that a document might no longer match a given query if we go from 4 to 16 because the numeric term that would match in the former case simply doesn't exist in the latter case.

(as a primer, the way old versions of Lucene index numbers is to convert them into a series of specially formatted _strings_ which are then indexed as terms. Newer versions of Lucene use a better system called `points` and the notion of precision step no longer applies).

rnewson: Noting there's a good chance that precision step is "just" about performance. the actual numeric value is indexed in all cases, but not all of the speed-up / lower precision terms are. This is all ancient history and poorly remembered (by me).

pgj: Thank you for the explanation.

This uplift would require a validation to eliminate the generic worries around the potential index format changes anyway.  Once we had the related end-to-end tests implemented, we can also experiment with the various configurations on precision step.

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
```diff
@@ -101,7 +101,9 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
     val fieldCache = FieldCache.DEFAULT
     var result = 0L
     for (cacheEntry <- fieldCache.getCacheEntries) {
-      result += RamUsageEstimator.sizeOf(cacheEntry.getValue())
+      val value = cacheEntry.getValue().asInstanceOf[Accountable]
```
rnewson: `CacheEntry`'s are not `Acccountable` so this line will throw an exception (perhaps the test suite doesn't exercise this code as it's behind a config toggle).

I think `RamUsageEstimator.shallowSizeOf` is the appropriate fix for the deprecated `sizeOf` method.

src/main/scala/com/cloudant/clouseau/ClouseauTypeFactory.scala
```diff
@@ -52,6 +52,22 @@ object ClouseauTypeFactory extends TypeFactory {
...
+  // wrapper is to help to preserve the original setting for
+  // compatibility reasons :-(
+  def maybeDoubleStored(stored: Store) = {
+    val result = new FieldType()
+    result.setIndexed(true)
+    result.setTokenized(true)
+    result.setOmitNorms(true)
+    result.setIndexOptions(IndexOptions.DOCS_ONLY)
+    result.setNumericType(FieldType.NumericType.DOUBLE)
+    result.setNumericPrecisionStep(4)
```
rnewson: I think we should do `8` here, since that's all we use in the range queries later.

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -302,27 +296,19 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
...
-                    drilldownQuery.add(new CategoryPath(category1: _*))
-                  } else { //if there are multiple values OR'd them, delete this else part after updating to Apache Lucene > 4.6
-                    val dim = category1(0)
-                    val categoryPaths: Array[CategoryPath] = new Array[CategoryPath](len - 1)
-                    for (i <- 1 until len) {
-                      categoryPaths(i - 1) = new CategoryPath(Array(dim, category1(i)): _*)
-                    }
-                    drilldownQuery.add(categoryPaths: _*)
+                val dim = category.head
+                for (path <- category.tail) {
```
rnewson, line 302: Is this the same effect? It appears to loop over each item in the tail and add each of them to the drilldownQuery. whereas the original code only calls drilldownQuery.add once.

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -633,7 +605,8 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs]) extends Service(ctx) w
     case e: ParseException =>
       ('error, ('bad_request, e.getMessage))
     case e =>
-      ('error, e.getMessage)
+      val stackTrace = (for (t <- e.getStackTrace()) yield t.toString()).mkString("\n")
+      ('error, (e.getClass, e.getMessage, stackTrace))
```
rnewson: why this change?

rnewson (changes requested):

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
pgj: It is the value stored inside a `CacheEntry` for which the type cast is called.  In [the definition of the `CacheEntry` class](https://github.com/apache/lucene/blob/releases/lucene-solr/4.10.4/lucene/core/src/java/org/apache/lucene/search/FieldCache.java#L701-L766), the value is passed as an `Accountable` at the constructor but the `getValue()` method returns an `Object`.  Nothing in the class definition implies a change in the type, that is why I thought it could be safely casted "back".

src/main/scala/com/cloudant/clouseau/IndexService.scala
pgj: The Facet API has changed, it is not possible to call `DrillDownQuery/add` with the full list any more, c.f. [4.6.1](https://github.com/apache/lucene/blob/releases/lucene-solr/4.6.1/lucene/facet/src/java/org/apache/lucene/facet/search/DrillDownQuery.java#L131-L155) vs. [4.10.4](https://github.com/apache/lucene/blob/releases/lucene-solr/4.10.4/lucene/facet/src/java/org/apache/lucene/facet/DrillDownQuery.java#L140-L152).  Note that there is a loop inside the 4.6.1 version.  It uses the following definition to determine the `dim`:

```java
String dim = paths[0].components[0];
```

and that is what I considered the `head` here, also in accordance with the removed `else` branch (the one with the note about the removal once leaving Lucene 4.6 behind).

src/main/scala/com/cloudant/clouseau/IndexService.scala
pgj: Sometimes a `NullPointerException` was thrown and it was very hard to find the source of that because it did not even have an associated message.  It helped me the work on this change, and I felt it could only do good if it is left there as is.

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
pgj: I tried to blow up the code but I could not.  I activated the `clouseau.field_cache.size` metric and read its value through the JMX interface (with the help of JConsole) but it was always zero.  I ran `make mango-test` and `make elixir-search` from CouchDB, and other metrics, such as `lru.misses`, `lru.evictions`, and `opens` did show values greater than zero.

I tend to believe this happens because `CacheEntry/getValue()` always returns `null`.

pgj: Ah, even better: the length of the array of the field cache entries itself is zero.

rnewson: my bad on this. I glazed over the 'getValue()' call which returns an Object which is also Accountable. 

Please add a test for whether the result of getValue is Accountable (so we don't blow up in a future change) and I think this section is fine.

pgj: It was eventually decided to remove this metrics all together, see https://github.com/cloudant-labs/clouseau/pull/76 .

src/main/scala/com/cloudant/clouseau/IndexService.scala
pgj: This one breaks the message format, so I decided to remove it.

pgj:
I am closing now in favor of #80.  I will leave the changes around and we can revisit the uplift later.

PR #75, opened 2023-05-21
fix: wildcards in numeric range queries

pgj:
When the `*` (wildcard) symbol is used in ranges, the parsing dies with a `NullPointerException`.  That is because input strings with the `null` value are not handled -- they should be treated as infinity with the respective sign (depending on the side of range where the wildcard is used).

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
+++ b/src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
@@ -61,12 +61,14 @@ class ClouseauQueryParser(version: Version,
                              upper: String,
                              startInclusive: Boolean,
                              endInclusive: Boolean): Query = {
-    if (isNumber(lower) && isNumber(upper)) {
-      NumericRangeQuery.newDoubleRange(field, 8, lower.toDouble,
-        upper.toDouble, startInclusive, endInclusive)
+    val lb = Option(lower).getOrElse("-Infinity")
+    val ub = Option(upper).getOrElse("Infinity")
+    if (isNumber(lb) && isNumber(ub)) {
+      NumericRangeQuery.newDoubleRange(
+        field, 8, lb.toDouble, ub.toDouble, startInclusive, endInclusive)
     } else {
       setLowercaseExpandedTerms(field)
-      super.getRangeQuery(field, lower, upper, startInclusive, endInclusive)
+      super.getRangeQuery(field, lb, ub, startInclusive, endInclusive)
     }
   }
 
--- a/src/test/scala/com/cloudant/clouseau/ClouseauQueryParserSpec.scala
+++ b/src/test/scala/com/cloudant/clouseau/ClouseauQueryParserSpec.scala
@@ -41,6 +41,14 @@ class ClouseauQueryParserSpec extends SpecificationWithJUnit {
       parser.parse("foo:[1.0 TO 2.0]") must haveClass[NumericRangeQuery[JDouble]]
     }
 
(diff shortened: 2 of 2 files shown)


jaydoane (approved):
Unit tests pass. Also tested with `make mango-test` and `make elixir-suite`.

src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
```diff
@@ -61,12 +61,14 @@ class ClouseauQueryParser(version: Version,
                              upper: String,
                              startInclusive: Boolean,
                              endInclusive: Boolean): Query = {
-    if (isNumber(lower) && isNumber(upper)) {
-      NumericRangeQuery.newDoubleRange(field, 8, lower.toDouble,
-        upper.toDouble, startInclusive, endInclusive)
+    val lb = Option(lower).getOrElse("-Infinity")
```
rnewson, line 64: this appears to expand the syntax to mean any input that the parser yielded null for will be a valid open range, which boxes us in for future changes and I think is unintentional?

rnewson: I would prefer to see the NPE handled. If it occurs in;

```
private def isNumber(str: String): Boolean = {
    fpRegex.matcher(str).matches()
  }
```

then we should just change isNumber to return false if str is null.

rnewson (changes requested):

src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
pgj: The `null` values in the second and third parameters of `QueryParser/getRangeQuery()` have a special meaning.  It denotes the start and end of open ranges as suggested by [the sources of Lucene 4.6.1](https://github.com/apache/lucene/blob/releases/lucene-solr/4.6.1/lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/QueryParser.java#L441):

```java
          boolean startOpen=false;
          boolean endOpen=false;
          if (goop1.kind == RANGE_QUOTED) {
            goop1.image = goop1.image.substring(1, goop1.image.length()-1);
          } else if ("*".equals(goop1.image)) {
            startOpen=true;
          }
          if (goop2.kind == RANGE_QUOTED) {
            goop2.image = goop2.image.substring(1, goop2.image.length()-1);
          } else if ("*".equals(goop2.image)) {
            endOpen=true;
          }
          q = getRangeQuery(field, startOpen ? null : discardEscapeChar(goop1.image), endOpen ? null : discardEscapeChar(goop2.image), startInc, endInc);
```

You can also see in [the `toString()` method of `TermRangeQuery`](https://github.com/apache/lucene/blob/releases/lucene-solr/4.6.1/lucene/core/src/java/org/apache/lucene/search/TermRangeQuery.java#L118-L132) that `null`s are uniquely translated back to stars:

```java
      buffer.append(includeLower ? '[' : '{');
      buffer.append(lowerTerm != null ? ("*".equals(Term.toString(lowerTerm)) ? "\\*" : Term.toString(lowerTerm))  : "*");
      buffer.append(" TO ");
      buffer.append(upperTerm != null ? ("*".equals(Term.toString(upperTerm)) ? "\\*" : Term.toString(upperTerm)) : "*");
      buffer.append(includeUpper ? ']' : '}');
```

rnewson: aha, great, thanks for digging in.

rnewson (approved):

PR #78, opened 2023-06-13
fix: release leaked index write lock on failure when closing

pgj:
When a database is deleted, the corresponding directory structure is removed entirely, even if the index writer has not yet been closed.  Due to the missing files, both attempts of closing the index and then rolling it back fail, which eventually leads to leak of resources, more specifically locks, due to [a known bug in Lucene versions prior to 4.7.1](https://issues.apache.org/jira/browse/LUCENE-5544).  In result, the subsequent search requests to the same index will fail when recreated.

As a minimal approach, work around the problem by pulling out the reference to the write lock and release it.


rnewson (approved):

rnewson (changes requested):
hm, the more I think the more I doubt. The real fix does this differently and releases other potentially open files.

pgj:
Close this one in favor of #80.

PR #79, opened 2023-06-15
chore: add metric on the count of NativeFSLocks held

pgj:
Lucene 4.6 is known to leak locks when an exception hits the rollback of an index writer.  Introduce an optional metric on the size of the internal store that tracks the currently held `NativeFSLock`s to be able to assess its severity in practice.

The method is specific to the Lucene internals and it uses Java reflection to extract the data as it is officially not accessible.

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory
 import scalang._
 import com.yammer.metrics.scala._
 import scala.collection.JavaConversions._
+import java.util.HashSet
 
 class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Service(ctx) with Instrumented {
 
@@ -94,6 +95,18 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
   val openTimer = metrics.timer("opens")
   val lru = new LRU()
   val waiters = Map[String, List[(Pid, Any)]]()
+  val countLocksEnabled = ctx.args.config.getBoolean("clouseau.count_locks", false)
+  if (countLocksEnabled) {
+    val lockClass = Class.forName("org.apache.lucene.store.NativeFSLock")
+    val field = lockClass.getDeclaredField("LOCK_HELD")
+    field.setAccessible(true)
+    val LOCK_HELD = field.get(null).asInstanceOf[HashSet[String]]
+    metrics.gauge("NativeFSLock.count")(getNativeFSLockHeldSize(LOCK_HELD))
+  }
+
+  def getNativeFSLockHeldSize(lockHeld: Collection[String]) = lockHeld.synchronized {
+    lockHeld.size
+  }
 
   override def handleCall(tag: (Pid, Any), msg: Any): Any = msg match {
     case OpenIndexMsg(peer: Pid, path: String, options: Any) =>


src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
```diff
@@ -94,6 +95,21 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
...
+  }
+
+  def getNativeFSLockHeldSize(): Long = {
+    synchronized {
+      val lockClass = Class.forName("org.apache.lucene.store.NativeFSLock")
+      val field = lockClass.getDeclaredField("LOCK_HELD")
+      field.setAccessible(true)
+      val LOCK_HELD = field.get(null).asInstanceOf[HashSet[String]]
+      field.setAccessible(false)
+      LOCK_HELD.size
```
rnewson: we need to hold a lock on the LOCK_HELD object (like everything else that access it), but this is really close.

pgj: Would this be an additional `synchronized` section?  Something like this:

```scala
  def getNativeFSLockHeldSize(): Long = {
    synchronized {
      val lockClass = Class.forName("org.apache.lucene.store.NativeFSLock")
      val field = lockClass.getDeclaredField("LOCK_HELD")
      field.setAccessible(true)
      val LOCK_HELD = field.get(null).asInstanceOf[HashSet[String]]
      field.setAccessible(false)
      LOCK_HELD.synchronized {
        LOCK_HELD.size
      }
    }
  }
```

rnewson: what was the reason for the original synchronized block you added?

pgj: I thought this to be needed because I am modifying the properties for the private static field and gauges could be pulled concurrently.  Otherwise it is just a read operation, which might not require explicit locking...?

pgj: Ah, I managed to come up with a simplified solution.  I have also got reminded that `synchronized` is needed for establishing the "happens before" guarantee on concurrent reads.

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
```diff
@@ -94,6 +95,19 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
...
+  if (countLocksEnabled) {
+    val lockClass = Class.forName("org.apache.lucene.store.NativeFSLock")
+    val field = lockClass.getDeclaredField("LOCK_HELD")
+    field.setAccessible(true)
+    val LOCK_HELD = field.get(null).asInstanceOf[HashSet[String]]
+    field.setAccessible(false)
+    metrics.gauge("NativeFSLock.count")(getNativeFSLockHeldSize(LOCK_HELD))
+  }
+
+  def getNativeFSLockHeldSize(LOCK_HELD: HashSet[String]) = LOCK_HELD.synchronized {
```
rnewson: A couple of non-blocking style nits (given the hopefully short duration this code is even needed);

1) function parameters should be snakecase (so `lockHeld`, say)
2) We only care that the value has a size method, so type of `Collection[String]` (or `Set[String]`).

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
rnewson: it's more than that. from the HashSet javadocs

> Note that this implementation is not synchronized. If multiple threads access a hash set concurrently, and at least one of the threads modifies the set, it must be synchronized externally.

We are just reading, but Lucene will be writing. If we don't synchronize our usage we can cause an exception when Lucene calls it, introducing a problem that doesn't exist otherwise.

rnewson (approved):
remember that https://github.com/cloudant/chef-repo/blob/master/cookbooks/collectd/files/default/clouseau.java will need enhancing to get this new value into the metric system.

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
```diff
@@ -94,6 +95,19 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
   val openTimer = metrics.timer("opens")
   val lru = new LRU()
   val waiters = Map[String, List[(Pid, Any)]]()
+  val countLocksEnabled = ctx.args.config.getBoolean("clouseau.count_locks", false)
+  if (countLocksEnabled) {
+    val lockClass = Class.forName("org.apache.lucene.store.NativeFSLock")
+    val field = lockClass.getDeclaredField("LOCK_HELD")
+    field.setAccessible(true)
+    val LOCK_HELD = field.get(null).asInstanceOf[HashSet[String]]
+    field.setAccessible(false)
```
rnewson: can we still call `size` on it if you make the private field inaccessible again?

pgj: Yes, I was able to access `size` after marking the field inaccessible.

rnewson (approved):

PR #83, opened 2023-11-23
Introduce types to supported analyzers

iilyak:
Refactor `createAnalyzer` to remove polymorphic arguments and use `Map[String, Any]`

The polymorphism makes it harder to introduce types. Because it requires union types. Which are not natively supported by the version of Scala we use. Elimination of the polymorphism would allow us to introduce types.

The introduction of types is the vehicle to solve type erasure problem we would have to deal with when we upgrade Scala to the next version.

The refactoring is done using following steps

1. Add type specific constructors for `Analyzer`
   - `createAnalyzerFromString(name: String)`
   - `createAnalyzer(options: Map[String, Any])`
   - `createAnalyzerFromStringInt(name: String)`
   - `createAnalyzerInt(options: Map[String, Any])`
2. Add type specific constructors for `OpenIndexMsg`
   - `fromMap(peer: Pid, path: String, map: Map[_, _])`
   - `fromAnalyzerName(peer: Pid, path: String, name: String)`
   - `fromKVsList(peer: Pid, path: String, options: List[_])`
3. Make sure we correctly go from `Any` to the concrete type
   - ```scala
      def collectKVs(list: List[_]): List[(String, Any)] =
          list.collect {
              case t @ (_: String, _: Any) => t
          }.asInstanceOf[List[(String, Any)]]
     ```
   - `options.get("name").map(_.asInstanceOf[String])`

# Assumptions

1. The keys of `options` passed to `OpenIndexMsg` are strings
2. It is ok to just ignore all non-string keys in `options`
3. The analyzer name is always a `String`
4. The `stopwords` is a list
5. The elements of a `stopwords` are strings
6. It is ok to skip non-strings elements of `stopwords`
7. The `fields` value is a list
8. The elements of a `fields` list are tuples `(String, String)`
9. It is ok to skip elements which do not match the `(String, String)` pattern
10. The `config` is a String in `('analyze, config, text)` message in AnalyzerService.handleCall and it should really be named `('analyze, analyzerName, text)`.

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/AnalyzerService.scala
+++ b/src/main/scala/com/cloudant/clouseau/AnalyzerService.scala
@@ -25,8 +25,8 @@ class AnalyzerService(ctx: ServiceContext[ConfigurationArgs]) extends Service(ct
   val logger = LoggerFactory.getLogger("clouseau.analyzer")
 
   override def handleCall(tag: (Pid, Any), msg: Any): Any = msg match {
-    case ('analyze, analyzerConfig: Any, text: String) =>
-      SupportedAnalyzers.createAnalyzer(analyzerConfig) match {
+    case ('analyze, analyzerName: String, text: String) =>
+      SupportedAnalyzers.createAnalyzerFromString(analyzerName) match {
         case Some(analyzer) =>
           ('ok, tokenize(text, analyzer))
         case None =>
--- a/src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
+++ b/src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
@@ -367,16 +367,19 @@ object SupportedAnalyzers {
         case None =>
           fallbackAnalyzer
       }
+      // The Per-field analyzers are always strings
       var fieldMap: Map[String, Analyzer] = options.get("fields") match {
-        case Some(fields: List[(String, Any)]) =>
-          fields map { kv =>
-            createAnalyzerInt(kv._2) match {
-              case Some(fieldAnalyzer) =>
-                (kv._1, fieldAnalyzer)
-              case None =>
-                (kv._1, defaultAnalyzer)
-            }
-          } toMap
+        case Some(fields: List[_]) =>
+          // ignore all fields of non string type
+          fields.collect {
+            case (field: String, analyzerName: String) =>
+              createAnalyzerFromStringInt(analyzerName) match {
+                case Some(fieldAnalyzer) =>
+                  (field, fieldAnalyzer)
+                case None =>
+                  (field, defaultAnalyzer)
+              }
+          }.toMap
         case _ =>
           Map.empty
       }
--- a/src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
+++ b/src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
@@ -116,241 +116,241 @@ object SupportedAnalyzers {
         }
       })
     case "arabic" =>
-      options.get("stopwords") match {
-        case Some(stopwords: List[String]) =>
+      getStopWords(options) match {
+        case Some(stopwords) =>
           Some(new ArabicAnalyzer(IndexService.version, stopwords))
-        case _ =>
+        case None =>
           Some(new ArabicAnalyzer(IndexService.version))
       }
     case "bulgarian" =>
-      options.get("stopwords") match {
-        case Some(stopwords: List[String]) =>
+      getStopWords(options) match {
+        case Some(stopwords) =>
           Some(new BulgarianAnalyzer(IndexService.version, stopwords))
-        case _ =>
+        case None =>
           Some(new BulgarianAnalyzer(IndexService.version))
       }
     case "brazilian" =>
-      options.get("stopwords") match {
-        case Some(stopwords: List[String]) =>
+      getStopWords(options) match {
+        case Some(stopwords) =>
           Some(new BrazilianAnalyzer(IndexService.version, stopwords))
-        case _ =>
+        case None =>
           Some(new BrazilianAnalyzer(IndexService.version))
       }
     case "catalan" =>
-      options.get("stopwords") match {
-        case Some(stopwords: List[String]) =>
+      getStopWords(options) match {
+        case Some(stopwords) =>
           Some(new CatalanAnalyzer(IndexService.version, stopwords))
-        case _ =>
+        case None =>
           Some(new CatalanAnalyzer(IndexService.version))
       }
     case "cjk" =>
-      options.get("stopwords") match {
-        case Some(stopwords: List[String]) =>
+      getStopWords(options) match {
+        case Some(stopwords) =>
           Some(new CJKAnalyzer(IndexService.version, stopwords))
-        case _ =>
+        case None =>
           Some(new CJKAnalyzer(IndexService.version))
       }
     case "chinese" =>
-      options.get("stopwords") match {
-        case Some(stopwords: List[String]) =>
+      getStopWords(options) match {
+        case Some(stopwords) =>
           Some(new SmartChineseAnalyzer(IndexService.version, stopwords))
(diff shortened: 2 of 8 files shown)


src/main/scala/com/cloudant/clouseau/AnalyzerService.scala
```diff
@@ -25,8 +25,8 @@ class AnalyzerService(ctx: ServiceContext[ConfigurationArgs]) extends Service(ct
   val logger = LoggerFactory.getLogger("clouseau.analyzer")
 
   override def handleCall(tag: (Pid, Any), msg: Any): Any = msg match {
-    case ('analyze, analyzerConfig: Any, text: String) =>
-      SupportedAnalyzers.createAnalyzer(analyzerConfig) match {
+    case ('analyze, analyzerName: String, text: String) =>
```
iilyak, line 28: One more assumption that it is always String.

iilyak:
I did verify my assumptions via reading code of:

- clouseau
- mango
- nouveau 
- dreyfus

As well as CouchDb and Cloudant documentation.

src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
```diff
@@ -122,248 +116,249 @@ object SupportedAnalyzers {
...
-        case _ =>
+        case None =>
           Some(new UAX29URLEmailAnalyzer(IndexService.version))
       }
     case "perfield" =>
       val fallbackAnalyzer = new StandardAnalyzer(IndexService.version)
-      val defaultAnalyzer: Analyzer = options.get("default") match {
-        case Some(defaultOptions) =>
-          createAnalyzerInt(defaultOptions) match {
+      // The "default" field is always a string
```
pgj, line 358: Unfortunately, this cannot always be assumed safely.  I have tried to run CouchDB's `mango-test` target for testing the changes with some `text` searches, and the [`16-index-selectors-test.IndexSelectorText.test_uses_partial_index_with_non_indexable_selector`](https://github.com/apache/couchdb/blob/eb1a5881bb536453a355f16dcb09b223cfaec41e/src/mango/test/16-index-selectors-test.py#L284) test caused an exception (as well as HTTP 500) at this point:

```console
[actor:13] ERROR overlock.threadpool.ErrorLoggedThread - Exception was thrown in thread actor:13
java.lang.ClassCastException: scala.collection.immutable.$colon$colon cannot be cast to java.lang.String
	at com.cloudant.clouseau.SupportedAnalyzers$$anonfun$1.apply(SupportedAnalyzers.scala:359)
	at com.cloudant.clouseau.SupportedAnalyzers$$anonfun$1.apply(SupportedAnalyzers.scala:359)
	at scala.Option.map(Option.scala:133)
	at com.cloudant.clouseau.SupportedAnalyzers$.createAnalyzerInt(SupportedAnalyzers.scala:359)
	at com.cloudant.clouseau.SupportedAnalyzers$.createAnalyzerInt(SupportedAnalyzers.scala:99)
	at com.cloudant.clouseau.SupportedAnalyzers$.createAnalyzer(SupportedAnalyzers.scala:81)
	at com.cloudant.clouseau.IndexService$.start(IndexService.scala:906)
	at com.cloudant.clouseau.IndexManagerService$$anonfun$handleCall$1$$anonfun$apply$1.apply$mcV$sp(IndexManagerService.scala:120)
	at com.cloudant.clouseau.IndexManagerService$$anonfun$handleCall$1$$anonfun$apply$1.apply(IndexManagerService.scala:120)
	at com.cloudant.clouseau.IndexManagerService$$anonfun$handleCall$1$$anonfun$apply$1.apply(IndexManagerService.scala:120)
	at com.yammer.metrics.scala.Timer.time(Timer.scala:17)
	at com.cloudant.clouseau.IndexManagerService$$anonfun$handleCall$1.apply(IndexManagerService.scala:119)
	at com.cloudant.clouseau.IndexManagerService$$anonfun$handleCall$1.apply(IndexManagerService.scala:118)
	at scalang.FunProcess$$anon$2.run(FunProcess.scala:60)
	at org.jetlang.core.BatchExecutorImpl.execute(BatchExecutorImpl.java:11)
	at org.jetlang.fibers.PoolFiber.flush(PoolFiber.java:82)
	at org.jetlang.fibers.PoolFiber.access$100(PoolFiber.java:21)
	at org.jetlang.fibers.PoolFiber$1.run(PoolFiber.java:38)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1152)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:622)
	at java.lang.Thread.run(Thread.java:748)
	at overlock.threadpool.ErrorLoggedThread.run(NamedThreadFactory.scala:40)
[actor:13] ERROR clouseau.main - Uncaught exception: scala.collection.immutable.$colon$colon cannot be cast to java.lang.String
```

The reason is that the analyzer passed over there is not a string but a list.  This is obviously a mistake, but note that the current code can somehow get away with it without raising an exception.  For the backwards compatibility, it should be understood how such cases are handled and assess if it is worth to maintain the respective old behavior.

pgj: Perhaps things like that could be prevented by implementing a validation on the Dreyfus (CouchDB) side and present the user with HTTP 400.

iilyak: I printed out options used to construct analyzer while running mango test suite

I saw only following "default" values
 
- keyword
- List(location)

However you are right even if it does look like a mistake we cannot just introduce strict validation.

src/main/scala/com/cloudant/clouseau/AnalyzerService.scala
rnewson: It doesn't have to be a string. While it's unlikely that someone would use _search_analyze with a complex definition, they can today.

rnewson:
The term "refactor" refers only to software edits that do not make functional changes, but just rearrange code to make a future functional change easier. The term has no value otherwise.

Please replace the word "refactor" with "change" or some synonym in the description, as clearly the PR makes many function changes in addition to the stated goal of "introducing types".

src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
```diff
@@ -87,23 +90,14 @@ object SupportedAnalyzers {
...
   }
 
-  def createAnalyzerInt(options: Any): Option[Analyzer] = options match {
-    case name: String =>
-      createAnalyzerInt(Map("name" -> name))
-    case list: List[(String, Any)] =>
-      try {
-        createAnalyzerInt(list.toMap)
-      } catch {
-        case e: ClassCastException => None
```
rnewson, line 97: this is important, I don't see it reintroduced elsewhere or made impossible to happen by other changes.

iilyak:
> Please replace the word "refactor" with "change" or some synonym in the description, as clearly the PR makes many function changes in addition to the stated goal of "introducing types".

The goal of this PR is to avoid public API changes. If it does change the protocol then it is a bug in the PR. I'll try harder to prevent the public API changes. Which probably would result in a very different version of this PR.

src/main/scala/com/cloudant/clouseau/SupportedAnalyzers.scala
iilyak: I did try to move this into the ClouseauTypeFactory and implement it as a partial function. The `collect` would skip things which do not match the types defined in the `case` check. So we fail faster and don't even go to type casting.

```
def collectKVs(list: List[_]): List[(String, Any)] =
    list.collect { case t @ (_: String, _: Any) => t }.asInstanceOf[List[(String, Any)]]
```

One of my assumptions turned out to be wrong. Specifically the assumption about `default` field to be a string. Which means I have to redo the PR.

iilyak:
closing in favor of https://github.com/cloudant-labs/clouseau/pull/84

PR #84, opened 2023-11-29
Introduce types to supported analyzers take2

iilyak:
Refactor `createAnalyzer` to remove polymorphic arguments and use `Map[String, Any]`

The polymorphism makes it harder to introduce types. Because it requires union types. Which are not natively supported by the version of Scala we use. Elimination of the polymorphism would allow us to introduce types.

The introduction of types is the vehicle to solve type erasure problem we would have to deal with when we upgrade Scala to the next version.

The refactoring is done using following steps

1. Introduce `AnalyzerOptions` class with type specific constructors
   - `def fromMap(map: Map[_, _])`
   - `def fromAnalyzerName(name: String)`
   - `def fromKVsList(options: List[_])`
2. Make sure we correctly go from `Any` to the concrete type. This PR uses the
      `.collect` combinator instead of relying on `ClassCastException`.

# Assumptions

1. The keys of `options` passed to `OpenIndexMsg` are strings.
2. It is ok to just ignore all non-string keys in `options`.
3. The analyzer name is either a `String` or a single `String` element wrapped
   in the `List`.
4. The `stopwords` is a list.
5. The elements of a stopwords are strings.
6. It is ok to skip non-strings elements of stopwords.
7. The fields value is a list.
8. The elements of a fields list are tuples.
   `(String, String)` or `(String, [String])`.
9. The config is a `String` in `('analyze, config, text)` message in
   `AnalyzerService.handleCall` and it should really be named
   `('analyze, analyzerName, text)`.

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/AnalyzerOptions.scala
+++ b/src/main/scala/com/cloudant/clouseau/AnalyzerOptions.scala
@@ -1,0 +1,42 @@
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not
+// use this file except in compliance with the License. You may obtain a copy of
+// the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations under
+// the License.
+
+package com.cloudant.clouseau
+
+class AnalyzerOptions private (inner: Map[String, Any]) {
+  def toMap = inner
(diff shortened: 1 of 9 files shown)


pgj:
In the description, items 1-2 and 4-5 seem to be the same, except for the formatting.

pgj:
The `mango-test` and `elixir-search` targets work with recent CouchDB `main`.  However, in the meantime I have fixed the previously mischievous Mango integration test case that tried to pass the default analyzer as a list of strings.

iilyak:
> In the description, items 1-2 and 4-5 seem to be the same, except for the formatting.

I removed duplication in the description of the PR

pgj (approved):
Great work @iilyak!

rnewson (approved):
nice improvement to type safety.

PR #85, opened 2023-12-08
Replace deprecated JavaConversions with JavaConverters

iilyak:
The `JavaConversions` are [deprecated in 2.12.0](https://www.scala-lang.org/api/2.12.1/scala/collection/JavaConversions$.html) (in favor of `JavaConverters`) and removed in 2.13.x.

The `JavaConverters` still is a temporary solution because it is also [deprecated](https://www.scala-lang.org/api/2.13.8/scala/collection/JavaConverters$.html) since 2.13.0 (in favor of `scala.jdk.CollectionConverters` ).

However we cannot go directly to `scala.jdk.CollectionConverters` because it is not available in 2.9.1.


pgj (approved):
Excellent!

PR #87, opened 2024-01-16
Modernize scala syntax

iilyak:
This PR removes deprecated forms of Scala syntax as well as 

- try to use Option type whenever possible
- validates the types of inner elements in nested structures
- add test cases for parsing

The main goal of the PR is to prepare the codebase for Scala version upgrade.

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
+++ b/src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
@@ -107,7 +107,7 @@ class ClouseauQueryParser(version: Version,
     fpRegex.matcher(str).matches()
   }
 
-  private def setLowercaseExpandedTerms(field: String) {
+  private def setLowercaseExpandedTerms(field: String) = {
     getAnalyzer match {
       case a: PerFieldAnalyzer =>
         setLowercaseExpandedTerms(a.getWrappedAnalyzer(field))
@@ -116,7 +116,7 @@ class ClouseauQueryParser(version: Version,
     }
   }
 
-  private def setLowercaseExpandedTerms(analyzer: Analyzer) {
+  private def setLowercaseExpandedTerms(analyzer: Analyzer) = {
     setLowercaseExpandedTerms(!analyzer.isInstanceOf[KeywordAnalyzer])
   }
 
--- a/src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
+++ b/src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
@@ -107,7 +107,7 @@ class ClouseauQueryParser(version: Version,
     fpRegex.matcher(str).matches()
   }
 
-  private def setLowercaseExpandedTerms(field: String) = {
+  private def setLowercaseExpandedTerms(field: String): Unit = {
     getAnalyzer match {
       case a: PerFieldAnalyzer =>
         setLowercaseExpandedTerms(a.getWrappedAnalyzer(field))
@@ -116,7 +116,7 @@ class ClouseauQueryParser(version: Version,
     }
   }
 
-  private def setLowercaseExpandedTerms(analyzer: Analyzer) = {
+  private def setLowercaseExpandedTerms(analyzer: Analyzer): Unit = {
     setLowercaseExpandedTerms(!analyzer.isInstanceOf[KeywordAnalyzer])
   }
 
--- a/src/main/scala/com/cloudant/clouseau/ClouseauTypeFactory.scala
+++ b/src/main/scala/com/cloudant/clouseau/ClouseauTypeFactory.scala
@@ -120,49 +120,51 @@ object ClouseauTypeFactory extends TypeFactory {
     result
   }
 
-  private def addFields(doc: Document, field0: Any): Unit = field0 match {
-    case (name: String, value: String, options: List[(String, Any)]) =>
-      val map = options.toMap
-      constructField(name, value, toStore(map), toIndex(map), toTermVector(map)) match {
-        case Some(field) =>
-          map.get("boost") match {
-            case Some(boost: Number) =>
-              field.setBoost(toFloat(boost))
-            case None =>
-              'ok
-          }
-          doc.add(field)
-          if (isFacet(map) && value.nonEmpty) {
-            val fp = FacetIndexingParams.DEFAULT
-            val delim = fp.getFacetDelimChar
-            if (!name.contains(delim) && !value.contains(delim)) {
-              val facets = new SortedSetDocValuesFacetFields(fp)
-              facets.addFields(doc, List(new CategoryPath(name, value)).asJava)
+  private def addFields(doc: Document, field0: Any): Unit = {
+    field0 match {
+      case (name: String, value: String, options: List[_]) =>
+        val map = options.collect { case t @ (_: String, _: Any) => t }.asInstanceOf[List[(String, Any)]].toMap
+        constructField(name, value, toStore(map), toIndex(map), toTermVector(map)) match {
+          case Some(field) =>
+            map.get("boost") match {
+              case Some(boost: Number) =>
+                field.setBoost(toFloat(boost))
+              case None =>
+                'ok
             }
-          }
-        case None =>
-          'ok
-      }
-    case (name: String, value: Boolean, options: List[(String, Any)]) =>
-      val map = options.toMap
-      constructField(name, value.toString, toStore(map), Index.NOT_ANALYZED, toTermVector(map)) match {
-        case Some(field) =>
-          doc.add(field)
-        case None =>
-          'ok
-      }
-    case (name: String, value: Any, options: List[(String, Any)]) =>
-      val map = options.toMap
-      toDouble(value) match {
-        case Some(doubleValue) =>
-          doc.add(new DoubleField(name, doubleValue, toStore(map)))
-          if (isFacet(map)) {
-            doc.add(new DoubleDocValuesField(name, doubleValue))
-          }
-        case None =>
-          logger.warn("Unrecognized value: %s".format(value))
-          'ok
-      }
+            doc.add(field)
+            if (isFacet(map) && value.nonEmpty) {
+              val fp = FacetIndexingParams.DEFAULT
+              val delim = fp.getFacetDelimChar
+              if (!name.contains(delim) && !value.contains(delim)) {
+                val facets = new SortedSetDocValuesFacetFields(fp)
+                facets.addFields(doc, List(new CategoryPath(name, value)).asJava)
+              }
+            }
+          case None =>
+            'ok
+        }
+      case (name: String, value: Boolean, options: List[_]) =>
+        val map = options.collect { case t @ (_: String, _: Any) => t }.asInstanceOf[List[(String, Any)]].toMap
+        constructField(name, value.toString, toStore(map), Index.NOT_ANALYZED, toTermVector(map)) match {
+          case Some(field) =>
+            doc.add(field)
+          case None =>
+            'ok
+        }
+      case (name: String, value: Any, options: List[_]) =>
+        val map = options.collect { case t @ (_: String, _: Any) => t }.asInstanceOf[List[(String, Any)]].toMap
+        toDouble(value) match {
+          case Some(doubleValue) =>
+            doc.add(new DoubleField(name, doubleValue, toStore(map)))
+            if (isFacet(map)) {
+              doc.add(new DoubleDocValuesField(name, doubleValue))
+            }
+          case None =>
+            logger.warn("Unrecognized value: %s".format(value))
+            'ok
+        }
+    }
   }
 
   private def constructField(name: String, value: String, store: Store, index: Index, tv: TermVector): Option[Field] = {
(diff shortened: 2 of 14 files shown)


src/main/scala/com/cloudant/clouseau/ClouseauTypeFactory.scala
```diff
@@ -120,49 +120,51 @@ object ClouseauTypeFactory extends TypeFactory {
...
-          doc.add(field)
-          if (isFacet(map) && value.nonEmpty) {
-            val fp = FacetIndexingParams.DEFAULT
-            val delim = fp.getFacetDelimChar
-            if (!name.contains(delim) && !value.contains(delim)) {
-              val facets = new SortedSetDocValuesFacetFields(fp)
-              facets.addFields(doc, List(new CategoryPath(name, value)).asJava)
+  private def addFields(doc: Document, field0: Any): Unit = {
+    field0 match {
+      case (name: String, value: String, options: List[_]) =>
```
pgj: Should not we keep the parameter to the `List[A]` type here?

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
```diff
@@ -104,7 +104,7 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
     metrics.gauge("NativeFSLock.count")(getNativeFSLockHeldSize(LOCK_HELD.asScala))
   }
 
-  def getNativeFSLockHeldSize(lockHeld: Collection[String]) = lockHeld.synchronized {
+  def getNativeFSLockHeldSize(lockHeld: scala.collection.mutable.Set[String]) = lockHeld.synchronized {
```
pgj, line 107: Why a more generic type, something similar to `Collection[A]` could work here?

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -932,4 +914,171 @@ object IndexService {
...
+      case Some(extractor(value)) =>
+        Some(value)
+      case None =>
+        None
+    }
+  }
+
+  /**
+   * Returns the Option[value] of a `ranges` key from given map of options if it is in correct format,
+   * otherwise throws ParseException.
```
pgj: Does this have to be mentioned explicitly?  The `@throws` annotation already supplies this information.

src/main/scala/com/cloudant/clouseau/Utils.scala
```diff
@@ -29,4 +29,55 @@ object Utils {
...
+   *    Some(value)
+   *  case None =>
+   *    None
+   * }
+   * ```
+   * In this example, the `asListOfStrings` partial function will convert any list to a `List[String]`,
+   * and throw an exception if it encounters a non-string element.
+   */
+
+  def elementWise[Inner](pf: PartialFunction[Any, Inner], orElse: PartialFunction[(List[Any], Any), List[Inner]]): PartialFunction[Any, List[Inner]] = {
```
pgj: Nice abstraction!

src/test/scala/com/cloudant/clouseau/IndexServiceSpec.scala
```diff
@@ -25,10 +25,192 @@ import scalang.Pid
...
+    }
+
+    "return correct result for empty list include_fields" in {
+      IndexService.getIncludeFields(includeFields(List())) must beEqualTo(Some(Set("_id")))
+    }
+
+    "return correct result for List[String]" in {
+      var planets = List("Mercury", "Venus", "Earth", "Mars", "Jupiter", "Saturn", "Uranus", "Neptune")
+      IndexService.getIncludeFields(includeFields(planets)) must beEqualTo(Some(
+        Set("Mercury", "Venus", "Earth", "Mars", "Jupiter", "Saturn", "Uranus", "Neptune", "_id")
```
pgj, line 88: Can the order of elements be guaranteed in case of a `Set()`?

iilyak: Set looses information about the order. Therefore it doesn't matter in which order the planets are specified when Set is constructed using literal syntax. I could say the following and it would still work.

```scala
Set("_id", "Neptune", "Uranus", "Saturn", "Jupiter", "Mars", "Earth", "Venus", "Mercury")
```

The reason why Set is chosen is to be compatible with existent implementation see [here](https://github.com/cloudant-labs/clouseau/blob/master/src/main/scala/com/cloudant/clouseau/IndexService.scala#L290)

```scala
    val includeFields: Set[String] =
      request.options.getOrElse('include_fields, 'nil) match {
        case 'nil =>
          null
        case value: List[String] =>
          Set[String]() ++ ("_id" :: value).toSet
        case other =>
          throw new ParseException(other + " is not a valid include_fields query")
      }
```

src/main/scala/com/cloudant/clouseau/ClouseauTypeFactory.scala
iilyak: If we would have `List[A]` then newer compiler would be producing warnings like this

```
warning: non-variable type argument (String, Any) in type pattern List[(String, Any)] (the underlying of List[(String, Any)]) is unchecked since it is eliminated by erasure
    case (name: String, value: String, options: List[(String, Any)]) =>
                                                     ^
```

New versions of scala do not have ability to match on inner type of containers. It cannot distinguish the following two cases

```scala
   case strings: List[Strings] => ...
   case integers: List[Int] => ...
```

As a result Scala designers decided to introduce the warning. The `List[_]` signals to compiler that user is not trying to match on the type so it is fine use.

iilyak: This also applies to generics. For example I cannot do

```scala
def asListOf[T](input: List[Any]): List[T] = {
    input match {
        case list: List[T] => list.asInstanceOf[List[T]]
        case _ => throw new Exception("")
    } 
}
```

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
iilyak: The `Collection` is gone in new scala.

src/main/scala/com/cloudant/clouseau/ClouseauTypeFactory.scala
iilyak: Personally I would prefer to keep the type and find a way to disable the warning. Because in this case I know what I am doing. But I couldn't find the way to control the warning. I'll try again.

iilyak: BTW the following also not going to work.

```scala
input match {
    case list if (list.isInstanceOf[List[String]]) => ...
```

iilyak: I've found a way

```scala
      case (name: String, value: String, options: List[(String, Any) @unchecked]) => ...
```

iilyak: I'll add the types back and use `@unchecked` trick.

pgj: Thank you @iilyak for both the explanation and the fix.

src/main/scala/com/cloudant/clouseau/IndexManagerService.scala
pgj: Ah, okay.

src/test/scala/com/cloudant/clouseau/IndexServiceSpec.scala
pgj: Right, makes sense.

pgj:
Are you planning to update the CI as well?  Apparently it tries to build the code with old Java (and Scala) and it fails.

iilyak:
> Are you planning to update the CI as well? Apparently it tries to build the code with old Java (and Scala) and it fails.

This work doesn't make it possible to build full project with new Java/Scala. I just renovated syntax of `*.scala` files without addressing the dependencies.

Therefore updating CI is not needed.

src/main/scala/com/cloudant/clouseau/Utils.scala
```diff
@@ -29,4 +29,55 @@ object Utils {
     new BytesRef(string)
   }
 
+  /**
+   * The function `elementWise` maps over elements of a list and applies a partial function `pf`
```
pgj: Based on the description, "elementWise" may not be the name that suits the definition the most.  If this is intended to be refine a generic list to a specific list, that is more like a coercion.  Something like `coerceList`.

iilyak: how about?

- `refineElementsType`
- `refineInnerType`
- `ensureInnerType`
- `verifyInnerType`
- `ensureElementsType`
- `coerceElementsType`
- `coerceInnerType`

src/main/scala/com/cloudant/clouseau/IndexService.scala
iilyak: Fixed in aff504c6c0fcffcb599d18800d62aae05d6c9be2

src/main/scala/com/cloudant/clouseau/Utils.scala
pgj: Thanks for the suggestions — I like `ensureElementsType`.

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -932,4 +914,170 @@ object IndexService {
...
+      case Some(extractor(value)) =>
+        Some(value)
+      case None =>
+        None
+    }
+  }
+
+  /**
+   * Returns the Option[List[String]] of a given option key from a map of options if each element of a list is indeed a String.
+   * It also throws ParseException if passed value is not a List[_].
```
pgj: There is an extra redundant mention of `ParseException` (c.f. the `@throws` clause).

src/main/scala/com/cloudant/clouseau/IndexService.scala
```diff
@@ -932,4 +914,170 @@ object IndexService {
...
+   */
+
+  private[clouseau] def getIncludeFields(options: Map[Symbol, Any]): Option[Set[String]] = {
+    getListOfStringsOption(options, 'include_fields).map(value => Set[String]() ++ ("_id" :: value).toSet)
+  }
+
+  /**
+   * Returns the Option[List[Product2[Option[GroupName], List[Any]]]] of a `groups` key from given map of options
+   * if each element of a list of tuples with arity 2 and first element is a String.
+   * It also throws ParseException if passed value is not a List[_].
```
pgj: `ParseException` is already hinted in `@throws`.  Maybe you can move the description over there as well?

src/main/scala/com/cloudant/clouseau/IndexService.scala
iilyak: addressed in a12e954

src/main/scala/com/cloudant/clouseau/IndexService.scala
iilyak: addressed in a12e954

pgj (approved):
Well done @iilyak!  Looks good to me now.

src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
```diff
@@ -107,7 +107,7 @@ class ClouseauQueryParser(version: Version,
     fpRegex.matcher(str).matches()
   }
 
-  private def setLowercaseExpandedTerms(field: String) {
+  private def setLowercaseExpandedTerms(field: String) = {
```
rnewson: why avoid "procedural syntax" for all these functions where `Unit` is the appropriate result? Is it being removed in Scala 3?

src/main/scala/com/cloudant/clouseau/DistanceValueSource.scala
```diff
@@ -25,11 +25,11 @@ import java.util.Map
 This is lucene spatial's DistanceValueSource but with configurable
 x and y field names to better suit our existing API.
  */
-class DistanceValueSource(ctx: SpatialContext,
-                          lon: String,
-                          lat: String,
-                          multiplier: Double,
-                          from: Point)
+case class DistanceValueSource(ctx: SpatialContext,
```
rnewson, line 28: neat.

rnewson (approved):
lots of nice work here. a little puzzled at the commit that removes the procedural syntax that gets followed up by adding back in the `: Unit =` return type, but if it's a style thing or a future deprecation, fine.

src/main/scala/com/cloudant/clouseau/ClouseauQueryParser.scala
iilyak: The procedural syntax has been removed in https://github.com/scala/scala/pull/3076. I am not sure in which version of scala this landed though. I can find out the version.

rnewson: no need. that's a perfectly good reason for the change.

PR #92, opened 2024-08-01
Release 2.23.1

jaydoane:
Plan to rebase merge and add `2.23.1` tag with following message:
```
Clouseau 2.23.1

- Support OTP 26.

```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.cloudant</groupId>
   <artifactId>clouseau</artifactId>
-  <version>2.23.1</version>
+  <version>2.23.2-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


nickva (approved):

pgj:
Similarly to the Scalang case, I could ponder if a minor version bump is warranted here.  But I guess we can say that a patch-level bump is sufficiently enough.  Ideally, it should not cause any serious change in the behavior or there is no risk of that happening.

jaydoane:
> Similarly to the Scalang case, I could ponder if a minor version bump is warranted here. But I guess we can say that a patch-level bump is sufficiently enough. Ideally, it should not cause any serious change in the behavior or there is no risk of that happening.

I had also considered a larger bump, but just decided to keep it as simple as possible, and also consistent with previous version changes when there was only a change in a dependency.

PR #100, opened 2025-02-03
chore: import dependencies into the source tree

pgj:
There is no need to host many of the customized dependencies of Clouseau in separate repositories as well as produce packages for them separately.  They are not used by other applications and they come with a lot of unused parts that are better to be trimmed anyway.

The following trees are being added here:

- https://github.com/boundary/logula @ `logula_slf4j_2.9.1-2.1.3-p01`
- https://github.com/cloudant/JCTools @ 010666e9
- https://github.com/cloudant/overlock @ cf6d7f71
- https://github.com/cloudant/scalang @ 15c8e17b

From `JCTools`, only specific classes are included from the `jctools-core` module.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -200,3 +200,40 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+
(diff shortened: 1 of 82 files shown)


pgj:
Closing in favor of #103.

PR #103, opened 2025-02-07
chore: import Scalang and its dependencies into the source tree

pgj:
There is no need to host many of the customized dependencies of Clouseau in separate repositories as well as produce packages for them separately.  They are not used by other applications and they come with a lot of unused parts that are better to be trimmed anyway.

The following trees are being added here:

- https://github.com/boundary/logula @ `logula_slf4j_2.9.1-2.1.3-p01`
- https://github.com/cloudant/JCTools @ 010666e9
- https://github.com/cloudant/overlock @ cf6d7f71
- https://github.com/cloudant/scalang @ 15c8e17b

From `JCTools`, only specific classes are included from the `jctools-core` module.  At the same time, remove the dependency to Apache Tika that is commented out, because it is not used any longer.

Note that the Cloudant-custom fork of Apache Lucene, tagged as `4.6.1-cloudant1` is something that could not be easily included in the tree and as such it remains a separate module.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,6 @@
     <maven.compiler.target>1.6</maven.compiler.target>
     <scala.version>2.9.1</scala.version>
     <scala.plugin.version>3.2.0</scala.plugin.version>
-    <tika-version>1.0</tika-version>
     <slf4j.version>1.7.32</slf4j.version>
   </properties>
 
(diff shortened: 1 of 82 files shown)


pom.xml
```diff
@@ -17,8 +17,7 @@
     <maven.compiler.source>1.6</maven.compiler.source>
     <maven.compiler.target>1.6</maven.compiler.target>
     <scala.version>2.9.1</scala.version>
-    <scala.plugin.version>3.1.0</scala.plugin.version>
-    <tika-version>1.0</tika-version>
```
jaydoane, line 21: Is it worth mentioning that you also removed this unused dependency?

jaydoane (approved):
I ran `mvn clean install -U` and `make clouseau1` then ran the testy search suite:
```
264 passed, 8 skipped, 10 xfailed, 5 warnings in 491.69s (0:08:11) 
```

pom.xml
pgj: Yes, and probably I should place that into a separate commit.

jaydoane (approved):

PR #106, opened 2025-05-28
Use a different method to return version

pgj:
The `getImplementationVersion` method can return version information only if it is run from a class inside a JAR file.  Choose a different method for retrieving version information so that reporting the version (over the Erlang Distribution Protocol) could work when the code is run directly from the source code directory (via `mvn scala:run`).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -352,6 +352,15 @@
         <version>2.4</version>
       </extension>
     </extensions>
+    <resources>
+      <resource>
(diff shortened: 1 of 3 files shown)


jaydoane (approved):
Works for me locally:
```
(node1@127.0.0.1)2> clouseau_rpc:version().
{ok,<<"2.25.1-SNAPSHOT">>}
```
Nice fix!

PR #108, opened 2025-09-09
Fix NullPointerException in `recursivelyDelete`

jaydoane:
While testing the ZIO-based `3.0.0` I noticed the following stack trace in code that is shared between code bases:
```
2025-08-31T22:08:03.280Z [DEBUG][zio-fiber-866023575][Actor.scala:165][zio-logger] Fiber zio-fiber-866023575 did not handle an error Exception in thread "zio-fiber-866023575" java.lang.NullPointerException: null
	at java.base/java.lang.reflect.Array.getLength(Native Method)
	at scala.collection.ArrayOps$.foreach$extension(ArrayOps.scala:1321)
	at com.cloudant.ziose.clouseau.IndexCleanupService.recursivelyDelete(IndexCleanupService.scala:79)
	at com.cloudant.ziose.clouseau.IndexCleanupService.$anonfun$recursivelyDelete$1(IndexCleanupService.scala:80)
	at com.cloudant.ziose.clouseau.IndexCleanupService.$anonfun$recursivelyDelete$1$adapted(IndexCleanupService.scala:79)
	at scala.collection.ArrayOps$.foreach$extension(ArrayOps.scala:1324)
	at com.cloudant.ziose.clouseau.IndexCleanupService.recursivelyDelete(IndexCleanupService.scala:79)
	at com.cloudant.ziose.clouseau.IndexCleanupService.handleCast(IndexCleanupService.scala:37)
	at com.cloudant.ziose.scalang.Service.$anonfun$onMessage$6(Service.scala:480)
	at com.cloudant.ziose.core.AddressableActor.handleActorMessage(Actor.scala:198)
	at com.cloudant.ziose.core.AddressableActor.tryCatch(Actor.scala:223)
	at com.cloudant.ziose.core.AddressableActor.handleActorMessage(Actor.scala:198)
	at com.cloudant.ziose.core.AddressableActor.start.loop(Actor.scala:152)
	at com.cloudant.ziose.core.AddressableActor.start.loop(Actor.scala:157)
	at com.cloudant.ziose.core.AddressableActor.start.loop(Actor.scala:149)
	at com.cloudant.ziose.core.AddressableActor.start(Actor.scala:165)
```
This wraps the output of `listFiles` in an `Option` to return an empty `Array`, which is safe to iterate over.

### DIFF
--- a/src/main/scala/com/cloudant/clouseau/IndexCleanupService.scala
+++ b/src/main/scala/com/cloudant/clouseau/IndexCleanupService.scala
@@ -73,7 +73,8 @@ class IndexCleanupService(ctx: ServiceContext[ConfigurationArgs]) extends Servic
 
   private def recursivelyDelete(fileOrDir: File, deleteDir: Boolean): Unit = {
     if (fileOrDir.isDirectory) {
-      for (file <- fileOrDir.listFiles)
+      // Wrap file list in `Option` to handle `null` case
+      for (file <- Option(fileOrDir.listFiles).getOrElse(Array()))
         recursivelyDelete(file, deleteDir)
       if (deleteDir)
         fileOrDir.delete


iilyak-ibm (approved):
+1

rnewson:
if `.isDirectory` returned true then `listFiles` shouldn't return null. We had a recent production issue (https://github.com/cloudant/chef-repo/pull/20554/files#diff-995285b8155d8f402b15202b444d67ea6b9ed46322da90610432e621cf82b6a2R45 for those that can see it) where we hadn't granted clouseau an apparmor rule to read the root directory where indexes were stored, so the recursion failed with an NPE immediately.

so, while not the nicest exception, I wonder if by "fixing" it you are masking the real cause.

All that to say that listFiles returning null _is_ unexpected, and handling (by ignoring) might not be appropriate.

jaydoane:
> if `.isDirectory` returned true then `listFiles` shouldn't return null

I believe that's correct unless`.isDirectory` is true, but then the directory in question is immediately deleted in a different thread, prior to `listFiles` being executed. At least, that seems to be how I was able to generate the error in the 3.0.0 code base.

rnewson:
hm ok, if something else is cleaning up dirs and you're racing against them, fair enough. As long as we have test coverage that `_search_cleanup` actually removes unreferenced indexes (I believe this was recently added in the releng test suite).

jaydoane:
> As long as we have test coverage that `_search_cleanup` actually removes unreferenced indexes (I believe this was recently added in the releng test suite).

I confirmed that the functionality was added to testy, but it also seems to be failing  most recent non-aborted runs, e.g.
```
[2025-08-23T00:39:28.515Z] suites/search/test_clouseau.py::TestClouseau::test_top_level_search_index_deletion 
[2025-08-23T00:39:28.515Z] Fatal error: run() received nonzero return code 2 while executing!
[2025-08-23T00:39:28.515Z] 
[2025-08-23T00:39:28.515Z] Requested: ls /srv/search_index/shards/00000000-1fffffff/testy028-user1/testy_db_mqxhztnxpd.*.deleted.* 
[2025-08-23T00:39:28.515Z] Executed: /bin/bash -l -c "ls /srv/search_index/shards/00000000-1fffffff/testy028-user1/testy_db_mqxhztnxpd.*.deleted.* "
[2025-08-23T00:39:28.515Z] 
[2025-08-23T00:39:28.515Z] =============================== Standard output ===============================
[2025-08-23T00:39:28.515Z] 
[2025-08-23T00:39:28.515Z] ls: cannot access '/srv/search_index/shards/00000000-1fffffff/testy028-user1/testy_db_mqxhztnxpd.*.deleted.*': No such file or directory
[2025-08-23T00:39:28.515Z] 
[2025-08-23T00:39:28.515Z] ===============================================================================
[2025-08-23T00:39:28.515Z] 
[2025-08-23T00:39:28.515Z] Aborting.
[2025-08-23T00:39:28.515Z] FAILED
```
so not sure if that counts as coverage 🤷

pgj:
> I confirmed that the functionality was added to testy [..] so not sure if that counts as coverage 🤷

Direct consumers of this repository cannot access that set of tests and it would also be nice for us to catch related regressions before the build gets to `testy`.  You may argue that it is worth the investment to extend the coverage for this case for this line of releases of Clouseau given that soon it will be replaced with another one where the related (local) tests will be in place.

jaydoane:
> > I confirmed that the functionality was added to testy [..] so not sure if that counts as coverage 🤷
> 
> Direct consumers of this repository cannot access that set of tests and it would also be nice for us to catch related regressions before the build gets to `testy`. You may argue that it is worth the investment to extend the coverage for this case for this line of releases of Clouseau given that soon it will be replaced with another one where the related (local) tests will be in place.

I agree that a unit test would be nice to have to validate change, but I also think in this case it's safe to merge without, because this fix has been validated:
- On a private Cloudant CI
- In Cloudant production, having been incorporated into `clouseau-3.0.0-rc15`

And because this `master` branch is about to become replaced with `main` containing the new 3.0.0 codebase it may not make sense for me to spend time on such a test.

PR #109, opened 2025-11-25
chore: Update README.md

willholley:
* Remove reference to `ziose`.
 * Update benefits vs 2.x.
 * Add upgrade notes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ However, there are some notable changes to the deployment:
 
  * Java 21 is required.
  * `clouseau.ini` is replaced by [app.conf](./app.conf).
- * CouchDB 3.??? required.
+ * CouchDB 3.5.0 required.
 
 ## Dependency management
 
--- a/README.md
+++ b/README.md
@@ -2,14 +2,24 @@
 
 Expose Lucene features to CouchDB via Erlang RPC.
 
-## Clouseau 3.x
-
-Originally dubbed `ziose`, version `3.x` replaces the foundation of [`clouseau`](https://github.com/cloudant-labs/clouseau/tree/master) with
-a [`ZIO`](https://github.com/zio/zio) as an asynchronous scheduler. Additionally, this version depends on the Java interface provided by Erlang/OTP.
-
 ## Build status
 
 [![Build 3.x](https://github.com/cloudant-labs/clouseau/actions/workflows/build.yaml/badge.svg)](https://github.com/cloudant-labs/clouseau/actions/workflows/build.yaml)
+
+## Clouseau 3.x
+
+Version `3.x` replaces the foundation of [`clouseau`](https://github.com/cloudant-labs/clouseau/tree/master) with
+a [`ZIO`](https://github.com/zio/zio) as an asynchronous scheduler. This improves the stability and performance,
+enables Clouseau to run on modern JVMs and brings all non-Lucene dependencies up-to-date.
+
+## Upgrading from 2.x
+
+Clouseau 3.x is intended to be a drop-in replacement for 2.x in that it remains compatible with existing Clouseau 2.x indexes. 
+However, there are some notable changes to the deployment:
+
+ * Java 21 is required.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,14 +2,24 @@
...
-## Build status
+## Upgrading from 2.x
 
-[![Build 3.x](https://github.com/cloudant-labs/clouseau/actions/workflows/build.yaml/badge.svg)](https://github.com/cloudant-labs/clouseau/actions/workflows/build.yaml)
+Clouseau 3.x is intended to be a drop-in replacement for 2.x in that it remains compatible with existing Clouseau 2.x indexes. 
+However, there are some notable changes to the deployment:
+
+ * Java 21 is required.
+ * `clouseau.ini` is replaced by [app.conf](./app.conf).
+ * CouchDB 3.??? required.
```
jaydoane:
```suggestion
 * CouchDB 3.5.0 required.
```

jaydoane: It's likely it will also work with earlier versions like 3.4.x, but I don't see any reason to make it more complicated.

jaydoane (approved):
Thanks for the cleanup!

README.md
```diff
@@ -2,14 +2,24 @@
 
 Expose Lucene features to CouchDB via Erlang RPC.
 
+## Build status
+
+[![Build 3.x](https://github.com/cloudant-labs/clouseau/actions/workflows/build.yaml/badge.svg)](https://github.com/cloudant-labs/clouseau/actions/workflows/build.yaml)
+
 ## Clouseau 3.x
 
-Originally dubbed `ziose`, version `3.x` replaces the foundation of [`clouseau`](https://github.com/cloudant-labs/clouseau/tree/master) with
```
pgj, line 7: I am not sure if removing this part of the sentence makes any difference.  When it was written, it was considered that the name `ziose` will be left in the code and it provides some kind of an explanation for that name.  Of course, it is not elaborated what the name `ziose` actually means — maybe that is what we should clarify here instead.

willholley: I don't see any harm in leaving it in the code, but most consumers won't be reading the code, I think.

Perhaps the naming history is something to mention in a `CONTRIBUTING.md`?

README.md
```diff
@@ -2,14 +2,24 @@
...
-## Build status
+## Upgrading from 2.x
 
-[![Build 3.x](https://github.com/cloudant-labs/clouseau/actions/workflows/build.yaml/badge.svg)](https://github.com/cloudant-labs/clouseau/actions/workflows/build.yaml)
+Clouseau 3.x is intended to be a drop-in replacement for 2.x in that it remains compatible with existing Clouseau 2.x indexes. 
+However, there are some notable changes to the deployment:
+
+ * Java 21 is required.
+ * `clouseau.ini` is replaced by [app.conf](./app.conf).
+ * CouchDB 3.5.0 required.
```
willholley, line 22: @jaydoane is this because of the OTP version that clouseau has been tested against?

jaydoane: @willholley not specifically the OTP version, but more generally we haven't tested with any older versions. That said, dreyfus hasn't really changed that much (by design), so it will probably "just work" with older versions, but I just don't see any upside to recommending older code. YMMV.

pgj: > I don't see any harm in leaving it in the code, but most consumers won't be reading the code, I think.

The name of the main class is different, it is `com.cloudant.ziose.clouseau.Main`.

README.md
jaydoane: The reason I left that line in was because there are still numerous references to `ziose` in the code, and I thought it would add some context. Parenthetically, I did attempt to remove a lot of those by eliminating the seemingly unnecessary `com/cloudant/ziose` path segments in some deep directory paths, but that PR never landed.

jaydoane:
I was planning to make README.md changes to support https://github.com/cloudant-labs/clouseau/pull/110 but it would be nice to see this merged first to prevent conflicts, if possible.

PR #112, opened 2025-11-30
Stop building on every push

jaydoane:
When testing workflows, building the project on every push was convenient but now it's redundant for a normal PR-to-`main` based workflow.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,9 +1,6 @@
 name: Build 3.x
 
 on:
-  push:
-    branches:
-      - '*'
   pull_request:
     branches:
       - main


pgj:
I think it can be useful to have a feedback from the CI even if I pushed commits to a branch but I have not opened the PR for that one.  The build takes only 20 minutes at maximum, it is not that long.  On the other hand, if we wanted to save effort, perhaps the solution is to filter by path — changes to files in the repository that do not contribute to the release artifacts should not trigger the CI-based checks.

jaydoane:
> I think it can be useful to have a feedback from the CI even if I pushed commits to a branch but I have not opened the PR for that one.

Yes indeed, that's exactly why it currently has both on push and on pull_request: because until now there was no `main` to merge into. However, now that there is a `main` branch, every PR will induce duplicate CI runs, which is annoying, especially since tests are flaky.

> The build takes only 20 minutes at maximum, it is not that long.

But the odds of both builds succeeding is much lower with the flakiness, so this will be a serious impediment to making any changes to `main`.

> On the other hand, if we wanted to save effort, perhaps the solution is to filter by path — changes to files in the repository that do not contribute to the release artifacts should not trigger the CI-based checks.

I think the solution is this PR. It's trivial to make the CI run simply by opening a PR against `main`: this is how almost every other project works, and I don't understand why we want to be different?

pgj:
> now that there is a main branch, every PR will induce duplicate CI runs, which is annoying, especially since tests are flaky

Okay, that explains the motivation for this change.  With that said, I do not mind applying it.

> this is how almost every other project works, and I don't understand why we want to be different?

Ideally, we should identify the causes of the flakiness and address those.  If those were fixed, it would not matter much how often the CI runs.  But we are not there yet, I can agree.

pgj (approved):

PR #114, opened 2025-12-07
fix(`otp`): shorten the ping period for monitoring the CouchDB node

pgj:
It is not recommended to specify very long timeout when calling the `OtpNode#ping` method.  There is only a single attempt is made to connect to the remote node.  This is also a blocking call where it is not so useful to keep it waiting for a long time.


jiahuili430 (approved):
+1

pgj:
Closing this in favor of #174 (as more advanced version).

PR #116, opened 2025-12-15
fix(`ci`): make `timeout` configurable and run in the foreground

pgj:
On macOS, the `timeout` command that is shipped with `coreutils` (GNU port) tends to get stuck for running `mango-test`.  Possibly that is not a coincidence that there exists a [native version](https://github.com/denispol/darwin-timeout) which is recommended for use instead.

Make it possible for the developers to override the use of that via factoring it out into a `make` variable.  While there, pass the `--foreground` flag to work around the issue by forbidding the tool going to the background.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ else
 OWASP_NVD_DATA_DIR?=/usr/share/dependency-check/data
 endif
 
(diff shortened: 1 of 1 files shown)


jiahuili430 (approved):
+1

PR #118, opened 2025-12-15
fix(`ci`): mitigate flakiness

pgj:
Certain CI steps happen to be flaky and sometimes it might make sense to simply retry them to overcome intermittent meta-issues.  For that purpose, call the commands for those steps via the [`retry` GitHub action](https://github.com/marketplace/actions/retry-step).

In parallel to that, pin the upstream CouchDB version to that of the latest release (of the time of writing, i.e. 3.5.1), because `main` recently had regressions which suggested the idea that we shall base our CI on a version other than the bleeding edge, so that we would be somewhat more protected from their breakages.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -25,16 +25,36 @@ jobs:
           install: true
 
       - name: Unit tests
-        run: mise exec -- make test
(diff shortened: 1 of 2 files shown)


jiahuili430 (approved):
+1

PR #119, opened 2025-12-15
Restore object form of fields in perfield analyzer definition

rnewson:
An analyzer, even one nested inside the "default" or "fields" section of a "perfield" analyzer can be expressed as an object.

Without this fix indexes have silently used StandardAnalyzer when users have declared other analyzers.

This significant regression was introduced in
https://github.com/cloudant-labs/clouseau/pull/84 and specifically this false assumption "The analyzer name is either a String or a single String element wrapped in the List."

### DIFF
--- a/clouseau/src/test/scala/com/cloudant/ziose/clouseau/SupportedAnalyzersSpec.scala
+++ b/clouseau/src/test/scala/com/cloudant/ziose/clouseau/SupportedAnalyzersSpec.scala
@@ -125,6 +125,25 @@ class SupportedAnalyzersSpec extends JUnitRunnableSpec {
           containsString("foo -> org.apache.lucene.analysis.en.EnglishAnalyzer")
         )
       },
+      test("perfield, field in object form") {
+        val options = {
+          AnalyzerOptions.from(Map("name" -> "perfield", "default" -> "standard",
+            "fields" -> List("foo" -> List(("name", "english")))))
+        }
+        assert(options)(isSome) &&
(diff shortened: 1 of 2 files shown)


James1912 (approved):

pgj:
Please note that in #122 I have ported the missing unit tests from Clouseau 2.x so they could be extended accordingly in case you wanted to cover this change with tests.

clouseau/src/test/scala/com/cloudant/ziose/clouseau/SupportedAnalyzersSpec.scala
```diff
@@ -125,6 +125,24 @@ class SupportedAnalyzersSpec extends JUnitRunnableSpec {
           containsString("foo -> org.apache.lucene.analysis.en.EnglishAnalyzer")
         )
       },
+      test("perfield, field in object form") {
+        val options = {
+          AnalyzerOptions.from(Map("name" -> "perfield", "default" -> "standard", "fields" -> List("foo" -> List(("name", "english")))))
```
pgj: This line is too long.  Please run `scalafmt` to wrap it.  Note that we have the `check-fmt` `make` target but apparently it is not run and therefore not enforced by the CI.

rnewson: done

pgj (approved):

PR #124, opened 2026-01-09
fix(`clouseau-ctrl`): handle when no node name could be discovered

pgj:
It is easy to forget about setting the node name for Clouseau on invoking the `clouseau-ctrl` utility, and the respective help message does not either imply that it is mandatory.  At the same time, once it is forgotten, the resulting error (with a stack trace) is not so user-friendly.

Add a pre-check for the node name before making any attempt to use that and bail out if it is still undefined.

### DIFF
--- a/clouseau-ctrl/src/clouseau_ctrl.erl
+++ b/clouseau-ctrl/src/clouseau_ctrl.erl
@@ -163,6 +163,10 @@ create_ctx(#{
     script_name := ScriptName,
     format := Format
 }) ->
+    case Name of
+        undefined -> fail("No node name has been specified!", []);
+        _ -> ok
+    end,
     % elp:ignore atoms_exhaustion - Atom exhastion is not an issue for CLIs (invoked once)
     RemoteNode = list_to_atom(Name ++ "@" ++ Host),
     % elp:ignore atoms_exhaustion - Atom exhastion is not an issue for CLIs (invoked once)


jiahuili430 (approved):
- main branch

```bash
$ make bin/clouseau_ctrl
$ bin/clouseau_ctrl service list -name clouseau1
-------------------------------------------------------------------------
|                        id                        |        pid         |
-------------------------------------------------------------------------
|                        init                      |     <9527.5.0>     |
|                        rex                       |     <9527.6.0>     |
|                      cleanup                     |     <9527.3.0>     |
|                        main                      |     <9527.2.0>     |
|                      analyzer                    |     <9527.4.0>     |
-------------------------------------------------------------------------

$ bin/clouseau_ctrl service list
escript: exception error: bad argument
  in operator  ++/2
     called as undefined ++ "@127.0.0.1"
  in call from clouseau_ctrl:create_ctx/1 (/Users/jiahuili/src/clouseau/clouseau-ctrl/src/clouseau_ctrl.erl, line 167)
  in call from clouseau_ctrl:init/1 (/Users/jiahuili/src/clouseau/clouseau-ctrl/src/clouseau_ctrl.erl, line 154)
  in call from clouseau_ctrl:service_list/1 (/Users/jiahuili/src/clouseau/clouseau-ctrl/src/clouseau_ctrl.erl, line 26)
  in call from escript:run/2 (escript.erl, line 750)
  in call from escript:start/1 (escript.erl, line 277)
  in call from init:start_em/1
  in call from init:do_boot/3
```

- After the change
```
$ bin/clouseau_ctrl service list
[failure] No node name has been specified!
```

Thank you for adding the pre-check function.
+1

PR #126, opened 2026-01-15
upgrade to Lucene 10.x

rnewson:
all tests pass


pgj:
> all tests pass

but one (see the RPC test (zeunit) logs):

```
  module 'index_tests'
    Test IndexService
      index_tests:108: -with/1-fun-0- (t_metrics_format)...*failed*
  in function index_tests:t_metrics_format/1 (/home/runner/work/clouseau/clouseau/zeunit/test/index_tests.erl, line 100)
  in call from eunit_test:run_testfun/1 (eunit_test.erl, line 71)
  in call from eunit_proc:run_test/1 (eunit_proc.erl, line 543)
  in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 368)
  in call from eunit_proc:handle_test/2 (eunit_proc.erl, line 526)
  in call from eunit_proc:tests_inorder/3 (eunit_proc.erl, line 468)
  in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 358)
  in call from eunit_proc:run_group/2 (eunit_proc.erl, line 582)
  **error:{assertNonNegInteger_failed,
      [{module,index_tests},
       {line,100},
       {expression,
           "maps : get ( 'partition_search.timeout.count' , Metrics , { no_such_key , 'partition_search.timeout.count' } )"},
       {pattern,"__V when is_integer ( __V ) andalso __V >= 0"},
       {value,{no_such_key,'partition_search.timeout.count'}}]}
    output:<<"">>
      [done in 0.005 s]
    [done in 0.313 s]
```

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -98,11 +94,9 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
   val deleteTimer = metrics.timer("deletes")
   val commitTimer = metrics.timer("commits")
 
-  val parSearchTimeOutCount = metrics.counter("partition_search.timeout.count")
```
pgj, line 101: Is this intended?  I think this is why the `zeunit` test on grabbing the metrics fails currently.

rnewson: fixed

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauQueryParser.scala
```diff
@@ -64,10 +64,16 @@ class ClouseauQueryParser(version: Version,
     val lb = Option(lower).getOrElse("-Infinity")
     val ub = Option(upper).getOrElse("Infinity")
     if (isNumber(lb) && isNumber(ub)) {
-      NumericRangeQuery.newDoubleRange(
-        field, 8, lb.toDouble, ub.toDouble, startInclusive, endInclusive)
+      var lv = lb.toDouble
```
pgj, line 67: Why do we have those `var`s here?  Is mutability a concern here?

jiahuili430: Maybe we can use this instead?

```scala
lazy val lv: Double = if (!startInclusive) Math.nextUp(lv) else lb.toDouble
lazy val uv: Double = if (!endInclusive) Math.nextDown(uv) else ub.toDouble
```

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauQueryParser.scala
```diff
@@ -76,48 +82,27 @@ class ClouseauQueryParser(version: Version,
...
                              quoted: Boolean): Query = {
     if (!quoted && isNumber(queryText)) {
-      new TermQuery(Utils.doubleToTerm(field, queryText.toDouble))
+      DoubleField.newExactQuery(field, queryText.toDouble)
     } else {
       super.getFieldQuery(field, queryText, quoted)
     }
   }
 
-  override def getFuzzyQuery(field: String, termStr: String,
```
jiahuili430, line 85: Why are we removing them?
`MultiFieldQueryParser.java` still have them, see
https://github.com/apache/lucene/blob/releases/lucene/10.3.2/lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/MultiFieldQueryParser.java#L207

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
```diff
@@ -77,83 +80,74 @@ object ClouseauTypeFactory extends TypeFactory {
         Some(Group2Msg(options.map(adapter.toScala).asInstanceOf[List[(Symbol, Any)]].toMap))
       }
       case ETuple(EAtom("update"), id: EBinary, fields: EList) => { // TODO verify maybe it should be EBinary(id)
+        val dvs = SMap[Tuple2[String,Symbol], IndexableField]()
         var doc = new Document()
```
jiahuili430, line 84: Is it okay to use `val` here?

clouseau/src/main/scala/com/cloudant/ziose/clouseau/DistanceValueSource.scala
```diff
@@ -30,44 +29,46 @@ case class DistanceValueSource(ctx: SpatialContext,
...
-    val reader: AtomicReader = readerContext.reader
+  val nullValue = 180 * multiplier
 
-    val ptLon: FieldCache.Doubles = FieldCache.DEFAULT.getDoubles(reader,
-      lon, true)
-    val ptLat: FieldCache.Doubles = FieldCache.DEFAULT.getDoubles(reader,
-      lat, true)
-    val validLon: Bits = FieldCache.DEFAULT.getDocsWithField(reader, lon)
-    val validLat: Bits = FieldCache.DEFAULT.getDocsWithField(reader, lat)
+  def description() = "DistanceValueSource(%s)".format(from)
```
jiahuili430, line 36: Perhaps we can simplify it like this?

```scala
def description = s"DistanceValueSource($from)"
override def toString = description
```

I couldn't find any other places where `description()` is used, so it might just be...
```scala
override def toString = s"DistanceValueSource($from)"
```

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -645,10 +587,10 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
...
           ctx.args.queryParser.parse(query)
         case Some(p) =>
-          val q = new BooleanQuery();
-          q.add(new TermQuery(new Term("_partition", p)), Occur.MUST);
-          q.add(ctx.args.queryParser.parse(query), Occur.MUST);
-          q
+          val builder = new BooleanQuery.Builder()
+          .add(new TermQuery(new Term("_partition", p)), Occur.MUST)
+          .add(ctx.args.queryParser.parse(query), Occur.MUST)
+          builder.build
```
jiahuili430: Shall we add a blank line before this, or add some spaces before `.add(...)`?

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -823,20 +764,64 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
...
 
-  private def convertFacetNode(node: FacetResultNode): Any = {
-    val children = node.subResults.asScala.map { n => convertFacetNode(n) }.toList
-    (node.label.components.toList, node.value, children)
+  private def toDoubleRangeFacetCounts(fc: FacetsCollector, field: RangesLabel, rangeQueries: List[Product2[RangesLabel, RangesQuery]]) = {
+    val ranges = new ListBuffer[DoubleRange]
+    for (rangeQuery <- rangeQueries) {
+      ctx.args.queryParser.parseDoubleRange(rangeQuery._1, rangeQuery._2) match {
+        case Some(doubleRange) => ranges += doubleRange
+        case None =>
```
jiahuili430: Is it ok to leave this blank? Do we need `()` here?

clouseau/src/test/scala/com/cloudant/ziose/clouseau/ClouseauQueryParserSpec.scala
```diff
@@ -19,8 +18,12 @@ import com.cloudant.ziose.test.helpers.TestRunner
...
 class ClouseauQueryParserSpec extends JUnitRunnableSpec {
   val TIMEOUT_SUITE = 5.minutes
-  val analyzer      = new StandardAnalyzer(IndexService.version)
-  def parser        = new ClouseauQueryParser(IndexService.version, "default", analyzer)
+  val analyzer      = new StandardAnalyzer()
+  def parser        = new ClouseauQueryParser("default", analyzer)
+
+  def isNumericRangeQuery(result: Any) = {
+    assert(result)(isSubtype[IndexOrDocValuesQuery](anything))
+  }
```
jiahuili430, line 26: Nice!

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
rnewson: I think so? dvs is a mutable map but we don't change dvs itself.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/DistanceValueSource.scala
rnewson: DistanceValueSource.scala is a cut-and-paste of DistanceValueSource.java except adapted to Scala and allowing us to choose the field names for longitude and latitude. You are right that description() is never called, but I wanted to make it obvious that this class is a close variant of the original.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauQueryParser.scala
rnewson: We only override them to do the `setLowercaseExpandedTerms` piece before delegating to the superclass, and that's no longer an option in Lucene 10. the prefix/fuzzy/etc query types still work though.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauQueryParser.scala
rnewson: They're `var` because they might be mutated immediately afterward. Jessica's online approach avoids that and is interesting, though I do not understand why we'd make them `lazy`.

jiahuili430: Without `lazy` compiler complains. I guess when the `if` statement is true, `lv` is not defined, so we need to use `lazy` here.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
jiahuili430: If `dvs.put(...)` works with `val`, then using `val` with the `val doc = new Document()` might also acceptable.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauQueryParser.scala
```diff
@@ -16,16 +16,13 @@ import java.util.regex.Pattern
...
-import org.apache.lucene.util.Version
-import org.apache.lucene.analysis.core.KeywordAnalyzer
+import org.apache.lucene.document.DoubleField
+import org.apache.lucene.facet.range.DoubleRange
 
-class ClouseauQueryParser(version: Version,
-                          defaultField: String,
+class ClouseauQueryParser(defaultField: String,
                           analyzer: Analyzer)
-    extends QueryParser(version, defaultField, analyzer) {
```
pgj, line 28: I wonder why the versions are now removed from everywhere (seemingly).

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
```diff
@@ -3,14 +3,16 @@ package com.cloudant.ziose.clouseau
...
-import org.apache.lucene.facet.params.FacetIndexingParams
-import org.apache.lucene.facet.sortedset.SortedSetDocValuesFacetFields
-import org.apache.lucene.facet.taxonomy.CategoryPath
+import org.apache.lucene.document.Field.Store
+import org.apache.lucene.document.{Document, StringField, TextField, DoubleDocValuesField, SortedDocValuesField}
 
-import scala.collection.JavaConverters._
 import org.apache.lucene.util.BytesRef
+import org.apache.lucene.document.DoublePoint
+import org.apache.lucene.document.StoredField
```
pgj, line 11: Minor nit: these two lines can be merged into line 7 (above).

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
```diff
@@ -77,83 +80,74 @@ object ClouseauTypeFactory extends TypeFactory {
...
-                    if (!name.contains(delim) && !value.contains(delim)) {
-                      val facets = new SortedSetDocValuesFacetFields(fp)
-                      facets.addFields(doc, List(new CategoryPath(name, value)).asJava)
-                    }
-                  }
-                case None =>
-                  'ok
+              if (value.nonEmpty) {
+                val map = options.collect { case t @ (key: String, value: Any) => t }.toMap
+                if (toIndex(map)) {
```
pgj, line 93: It might be rewritten as

```scala
val field = if (toIndex(map)) new Textfield(name, value, toStore(map)) else new StringField(name, value, toStore(map))
doc.add(field)
```

clouseau/src/main/scala/com/cloudant/ziose/clouseau/DistanceValueSource.scala
```diff
@@ -12,14 +12,13 @@
...
-import org.apache.lucene.util.Bits
-import com.spatial4j.core.distance.DistanceCalculator
-import java.util.Map
+import org.locationtech.spatial4j.context.SpatialContext
+import org.locationtech.spatial4j.shape.Point
+import org.apache.lucene.search.DoubleValuesSource
+import org.apache.lucene.index.LeafReaderContext
+import org.apache.lucene.search.DoubleValues
+import org.apache.lucene.index.DocValues
+import org.apache.lucene.search.IndexSearcher
```
pgj, line 21: Merge opportunity: lines 17, 19, and 21.  (They are all imported from the same package.)

clouseau/src/main/scala/com/cloudant/ziose/clouseau/DistanceValueSource.scala
```diff
@@ -12,14 +12,13 @@
...
-import org.apache.lucene.search.FieldCache
-import org.apache.lucene.util.Bits
-import com.spatial4j.core.distance.DistanceCalculator
-import java.util.Map
+import org.locationtech.spatial4j.context.SpatialContext
+import org.locationtech.spatial4j.shape.Point
+import org.apache.lucene.search.DoubleValuesSource
+import org.apache.lucene.index.LeafReaderContext
+import org.apache.lucene.search.DoubleValues
+import org.apache.lucene.index.DocValues
```
pgj, line 20: This line can be combined with line 18.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/DistanceValueSource.scala
```diff
@@ -30,44 +29,46 @@ case class DistanceValueSource(ctx: SpatialContext,
...
                                multiplier: Double,
                                from: Point)
-    extends ValueSource {
+    extends DoubleValuesSource {
 
-  def description() = "DistanceValueSource(%s)".format(from)
-
-  def getValues(context: Map[_, _], readerContext: AtomicReaderContext) = {
-    val reader: AtomicReader = readerContext.reader
+  val nullValue = 180 * multiplier
```
pgj, line 34: This does not look like a "null" value.  What is the purpose of this?

clouseau/src/main/scala/com/cloudant/ziose/clouseau/LoggerFactory.scala
```diff
@@ -81,8 +81,12 @@ object LoggerLayers {
 
 object LoggerFactory {
   case class NamedLogger(id: String) extends ZioSupport {
+    def isDebugEnabled()(implicit adapter: Adapter[_, _]): Boolean = {
+      LogLevel.Debug >= adapter.logLevel
+    }
+
     def debug(msg: => String)(implicit adapter: Adapter[_, _], trace: Trace): Unit = {
-      if (LogLevel.Debug >= adapter.logLevel) {
+      if (isDebugEnabled) {
```
pgj, line 89: I am not sure why this change was made.  I understand that it probably improves readability but it seems to be used at this single location only...?

clouseau/src/test/scala/com/cloudant/ziose/clouseau/IndexServiceSpec.scala
```diff
@@ -768,7 +767,7 @@ class IndexServiceSpec extends JUnitRunnableSpec {
                       List(
                         (
                           List("ffield"),
-                          0.0,
+                          3.0,
```
pgj, line 770: Why did this change?  (I assume the others below are also related.)

clouseau/src/main/scala/com/cloudant/ziose/clouseau/QueryExplainer.scala
```diff
@@ -1,92 +0,0 @@
...
-// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
-// License for the specific language governing permissions and limitations under
-// the License.
-
-package com.cloudant.ziose.clouseau
-
-import java.lang.{ StringBuilder, String }
-import org.apache.lucene.search._
-
-object QueryExplainer {
```
pgj, line 18: Why was this deleted?  Has this functionality been removed in Lucene 10 or we never used it?

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -66,8 +54,16 @@ import zio.ZIO
...
-
-case class IndexServiceArgs(config: Configuration, name: String, queryParser: QueryParser, writer: IndexWriter)
+import org.apache.lucene.facet.FacetsCollectorManager
+import org.apache.lucene.facet.FacetsCollector
+import org.apache.lucene.facet.DrillDownQuery
+import org.apache.lucene.facet.range.DoubleRangeFacetCounts
+import org.apache.lucene.facet.Facets
+import scala.collection.mutable.ListBuffer
+import org.apache.lucene.facet.sortedset.DefaultSortedSetDocValuesReaderState
+import org.apache.lucene.facet.sortedset.SortedSetDocValuesFacetCounts
```
pgj, line 64: There are a lot of opportunities to merge these import lines between many classes share the same modules.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -285,10 +289,6 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
       case e: AlreadyClosedException => 'ignored
       case e: IOException =>
         logger.warn(prefix_name("Error while closing writer"), e)
-        val dir = ctx.args.writer.getDirectory
-        if (IndexWriter.isLocked(dir)) {
```
pgj, line 289: Hrm, so no more locking is required for the index directories?

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -919,8 +903,9 @@ object IndexService {
     try {
       SupportedAnalyzers.createAnalyzer(options) match {
         case Some(analyzer) =>
-          val queryParser = new ClouseauQueryParser(version, "default", analyzer)
-          val writerConfig = new IndexWriterConfig(version, analyzer)
+          val queryParser = new ClouseauQueryParser("default", analyzer)
+          val writerConfig = new IndexWriterConfig(analyzer)
+          writerConfig.setUseCompoundFile(false)
```
pgj, line 908: Why did not you select the use of compound files?

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauQueryParser.scala
rnewson: because the classes no longer take this argument (https://github.com/apache/lucene/issues/6921)

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
rnewson: true, but that line existed before the lucene 10 upgrade. I try not to make unrelated changes in a PR just because they're adjacent.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
rnewson: hm, I don't much like that but it's a style not substance thing.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/DistanceValueSource.scala
rnewson: 🤷 https://github.com/apache/lucene/blob/2cbf261/lucene/spatial-extras/src/java/org/apache/lucene/spatial/vector/DistanceValueSource.java#L47

clouseau/src/main/scala/com/cloudant/ziose/clouseau/LoggerFactory.scala
rnewson: because I use isDebugEnabled elsewhere to avoid doing an expensive operation if debug is not enabled (which it almost always is not).

clouseau/src/test/scala/com/cloudant/ziose/clouseau/IndexServiceSpec.scala
rnewson: because now it is being counted correctly (this is the parent count, you can see its the sum of the previously tested facet counts below).

clouseau/src/main/scala/com/cloudant/ziose/clouseau/QueryExplainer.scala
rnewson: QueryExplainer isn't used in the product (and is not part of Lucene). I just didn't want to spend the time converting it to Lucene 10 when it's unused.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
rnewson: I let the IDE (or lack of) manage those things. we could do a separate PR that organises imports.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
rnewson: https://issues.apache.org/jira/browse/LUCENE-6060

Locking is required, but the ability to forcibly unlock is no longer available.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
rnewson: They are slower. compoundFile is really for Windows systems that struggle (or perhaps _struggled in the past_) with lots of file handles.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/LoggerFactory.scala
rnewson: at https://github.com/cloudant-labs/clouseau/pull/126/changes#diff-07b33dce862cc88a026f9ee880d4d8c2c343d62befeace8589e8b3ba797abda2R409

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
rnewson: indents added, thanks.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
rnewson: added ()

pgj (approved):
I think this version qualifies for the first release candidate build.  I could not find any evident issues with changes, all looks good to me now.  There are a couple of style-related suggestions pending still, but none of them is considered a blocker.

rnewson:
Ideally we'd have a tool that automatically formats the Java code and organises imports, rather than do it by hand. I'd like to defer those style nits for a separate PR

PR #127, opened 2026-01-17
chore: improve code formatting support

pgj:
Since the project uses a code format implemented by `scalafmt`, it might be useful to provide a dedicated `make(1)` target, `scalafmt-format` for developers to invoke the command.  This is created in line with the existing `erlfmt-format` that works with Erlang sources.

While here, add an extra `format-code` target that calls both `scalafmt-format` and `erlfmt-format` for convenience.  Also, extend the CI with calling the linters for running source code checks on the pull requests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,9 +130,9 @@ ERL_SRCS?=$(shell git ls-files -- "*/rebar.config" "*.[e,h]rl" "*.app.src" "*.es
 .PHONY: check-fmt
 # target: check-fmt - Check mis-formatted code
 check-fmt: $(ARTIFACTS_DIR)
-	@scalafmt --test | tee $(ARTIFACTS_DIR)/scalafmt.log
-	@ec | tee $(ARTIFACTS_DIR)/editor-config.log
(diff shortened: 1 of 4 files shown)


rnewson (approved):

Makefile
```diff
@@ -139,6 +139,15 @@ check-fmt: $(ARTIFACTS_DIR)
 erlfmt-format:
 	@erlfmt --write -- $(ERL_SRCS)
 
+.PHONY: scalafmt-format
+# target: scalafmt-format - Format Scala code automatically
+scalafmt-format:
+	@scalafmt --quiet
+
+.PHONY: format-code
```
jiahuili430, line 147: How about simplifying it to `fmt`?

pgj: It would not be consistent with other targets such as `erlfmt-format` and `scalafmt-format` so I think it is better keep it like that.

jiahuili430: As we discussed at the meeting, it's up to you.

PR #129, opened 2026-01-24
fix(`clouseau`): indexing for document fields of unrecognized types

pgj:
In 312656f3, a distinction between each supported type has been introduced but there was no case defined for values with unrecognized types.  This change broke the compatibility with the 2.x line of releases because such instances will make the process crash, while originally there should have been only a warning issued in the logs and the field in question should have been ignored.

Restore this case from the 2.x code base and while there, merge all the similar cases (about handling fields with various number types) since they share the same body.  They could be broken down to further cases later if there is a specific reason to do so.


jiahuili430 (approved):
+1

rnewson (approved):

PR #132, opened 2026-02-12
chore: track indexes seen

pgj:
Expose a metric that keeps track of the indexes opened during the last week, in a breakdown of the last minute, hour, and day.  Each index is considered only once.  This can be useful to understand how many indexes are actually being used, i.e. what the size of the respective working set is, and how to adjust the size of the LRU cache to reduce the chance of misses and evictions for the improved performance (at the cost of potentially higher memory consumption).

While here, amend the implementation of gauges because they were apparently broken.


clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexManagerService.scala
```diff
@@ -87,6 +94,23 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs])(implicit adapt
       }
     }
 
+    private val trackIndexesSeenRange = Duration.ofDays(7).toSeconds
```
rnewson, line 97: can we put this choice of value in config so ops can tune up (or more likely down) if they encounter issues?

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexManagerService.scala
```diff
@@ -99,12 +123,15 @@ class IndexManagerService(ctx: ServiceContext[ConfigurationArgs])(implicit adapt
...
+    private def now: Long = ChronoUnit.SECONDS.between(Instant.EPOCH, Instant.now)
+
   }
 
   val logger = LoggerFactory.getLogger("clouseau.main")
   val rootDir = new File(ctx.args.config.getString("clouseau.dir", "target/indexes"))
   val openTimer = metrics.timer("opens")
-  val lru = new LRU()
+  val trackIndexATimes = ctx.args.config.getBoolean("clouseau.track_index_atimes", false)
+  val lru = new LRU(trackIndexAccesses = trackIndexATimes)
```
rnewson, line 134: It looks like you intended the assignment here but it could confuse future readers. suggest you put the assignment on its own line and then this line is simply `val lru = new LRU(trackIndexAccesses)`.

rnewson (changes requested):
this is very good, I have a couple of comments. Making the duration we retain configurable is the main one.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexManagerService.scala
pgj: Sure, we could.  But if this is set to lower than 7 days it may cripple the metrics.  Perhaps I could make the exposure of the week-level metrics dependent on the fact that we store enough data for that — if we do not, then simply do not expose it.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexManagerService.scala
pgj: That was done for picking the third parameter for the class constructor and omit the other two (so that their default values will be assumed).

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexManagerService.scala
rnewson: ok, don't worry. Ops can always turn the whole thing off it it becomes a problem.

rnewson (approved):

PR #133, opened 2026-02-16
Test only

jiahuili430:


prometheus/JmxPrometheus.java
```diff
@@ -0,0 +1,35 @@
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.net.URI;
+import java.util.List;
+
+public class JmxPrometheus {
```
pgj: Cannot you just use `curl` to get the metrics and check the retrieved body with the usual command-line text processing UNIX tools?  It would be much simpler, I believe.

Makefile
```diff
@@ -178,6 +178,47 @@ jar: $(ARTIFACTS_DIR)/$(JAR_PROD)
...
+	@wget -nc $(JMX_EXPORTER_URL) -O $(JMX_EXPORTER) || true
+	@java -javaagent:$(JMX_EXPORTER)=$(JMX_EXPORTER_PORT):$(JMX_EXPORTER_CFG) \
+		-jar $(JAR_ARTIFACTS)
+
+prometheus/JmxPrometheus.class: prometheus/JmxPrometheus.java
+	javac -source 21 -target 21 "$<"
+
+.PHONY: prometheus-tests
+# target: prometheus-tests - Run JMX Prometheus metrics collection tests
+prometheus-tests: epmd $(JAR_ARTIFACTS) prometheus/JmxPrometheus.class FORCE
```
pgj: It might make sense to factor out the common parts from the `metrics-tests` and re-use them here?  Of course, `make` works a bit differently compared to other programming languages, but you may find examples of those in this file as well.

jiahuili430: This branch is used for running CI only. I will update `prometheus3` branch based on the feedback.
Thank you for the review and suggestions!

jiahuili430:
Used this branch to run CI, close this PR.

PR #139, opened 2026-03-09
fix grouping

rnewson:


pgj (approved):
Great find, thanks for the quick fix!

PR #143, opened 2026-03-13
Commit after setting the purge sequence

nickva:
This is a companion PR for Dreyfus [1]

The purge sequence value when set via the API is not committed, it's just stored in memory only until commit time. With the PR [1], if we update the checkpoint document on the db side, and purges get removed by the compactor, then, the index service also crashes before it commits, it's possible that the index would miss some purges. So to keep them in sync ensure that always commit after updating the purge sequence value. We'll first set the purge sequence in the index the update the checkpoint document. We'll make sure to call the set_purge_seq only when it gets updated or initialized and when the index is created (and initialized).

[1] https://github.com/apache/couchdb/pull/5920

### DIFF
--- a/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
+++ b/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
@@ -198,6 +198,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
       'ok
     case SetPurgeSeqMsg(newPurgeSeq: Long) =>
       pendingPurgeSeq = newPurgeSeq
+      commit(pendingSeq,  pendingPurgeSeq)
       logger.debug(prefix_name("purge sequence is now %d".format(newPurgeSeq)))
       'ok
     case 'info =>


clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -198,6 +198,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
       'ok
     case SetPurgeSeqMsg(newPurgeSeq: Long) =>
       pendingPurgeSeq = newPurgeSeq
+      commit(pendingSeq,  pendingPurgeSeq)
```
nickva, line 201: I am not sure if this will actually commit, it seems to spawn a separate process and we're not waiting for it in `commit/2` and if it's already `committing` it won't do anything. I don't think we have any "commit and return when done" facility we can use?

(I am not very familiar with Scala so these just random guesses of what's going on)

rnewson: that function spawns a process that does the commit (which could take some seconds) and then it sends a message back on completion or error. Maybe break out the bits that do the work into a commitSync function and call that

rnewson:
```diff
diff --git i/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala w/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
index dbcd6c26..1bd66f1d 100644
--- i/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
+++ w/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
@@ -83,7 +83,6 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
   var updateSeq = getCommittedSeq
   var pendingSeq = updateSeq
   var purgeSeq = getCommittedPurgeSeq
-  var pendingPurgeSeq = purgeSeq
   var committing = false
   var forceRefresh = false
   var idle = true
@@ -197,7 +196,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
       logger.debug(prefix_name("Pending sequence is now %d".format(newSeq)))
       'ok
     case SetPurgeSeqMsg(newPurgeSeq: Long) =>
-      pendingPurgeSeq = newPurgeSeq
+      commitSync(pendingSeq, newPurgeSeq)
       logger.debug(prefix_name("purge sequence is now %d".format(newPurgeSeq)))
       'ok
     case 'info =>
@@ -240,7 +239,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
       }
       exit('deleted)
     case 'maybe_commit =>
-      commit(pendingSeq, pendingPurgeSeq)
+      maybeAsyncCommit(pendingSeq, purgeSeq)
     case ('committed, newUpdateSeq: Long, newPurgeSeq: Long) =>
       updateSeq = newUpdateSeq
       purgeSeq = newPurgeSeq
@@ -295,18 +294,13 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
     ()
   }

-  private def commit(newUpdateSeq: Long, newPurgeSeq: Long) = {
+  private def maybeAsyncCommit(newUpdateSeq: Long, newPurgeSeq: Long) = {
     if (!committing && (newUpdateSeq > updateSeq || newPurgeSeq > purgeSeq)) {
       committing = true
       val index = self
       node.spawn(_ => {
-        ctx.args.writer.setCommitData((ctx.args.writer.getCommitData.asScala +
-          ("update_seq" -> newUpdateSeq.toString) +
-          ("purge_seq" -> newPurgeSeq.toString)).asJava)
         try {
-          commitTimer.time {
-            ctx.args.writer.commit()
-          }
+          commitSync(newUpdateSeq, newPurgeSeq)
           index ! ('committed, newUpdateSeq, newPurgeSeq)
         } catch {
           case e: AlreadyClosedException =>
@@ -320,6 +314,15 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
     }
   }

+  private def commitSync(newUpdateSeq: Long, newPurgeSeq: Long) = {
+    ctx.args.writer.setCommitData((ctx.args.writer.getCommitData.asScala +
+      ("update_seq" -> newUpdateSeq.toString) +
+      ("purge_seq" -> newPurgeSeq.toString)).asJava)
+    commitTimer.time {
+      ctx.args.writer.commit()
+    }
+  }
+
   private def search(request: SearchRequest): Any = {
     val queryString = request.options.getOrElse('query, "*:*").asInstanceOf[String]
     val refresh = request.options.getOrElse('refresh, true).asInstanceOf[Boolean]
```

PR #146, opened 2026-03-17
reply to SetPurgeSeqMsg after commit

rnewson:
We update the purge checkpoint doc after clouseau_rpc:set but this isn't safe as we've not committed.

on receipt of SetPurgeMsg start an async commit and delay the gen_server call reply till it happens.

### DIFF
--- a/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
+++ b/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
@@ -69,6 +69,7 @@ import zio.Duration
 
 case class IndexServiceArgs(config: Configuration, name: String, queryParser: QueryParser, writer: IndexWriter)
 case class HighlightParameters(highlighter: Highlighter, highlightFields: List[String], highlightNumber: Int, analyzers: List[Analyzer])
+case class CommitWaiter(tag: (Pid, Any), min_purge_seq: Long)
 
 // These must match the records in dreyfus.
 case class TopDocs(updateSeq: Long, totalHits: Long, hits: List[Hit])
@@ -88,6 +89,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
   var forceRefresh = false
   var idle = true
   var concurrentRequests = new AtomicInteger(0)
+  var commitWaiters = List[CommitWaiter]()
 
   def reader = lazyReader.getOrElse(throw InvalidReader)
   def setReader(reader: DirectoryReader) =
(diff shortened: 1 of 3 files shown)


clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -199,7 +201,13 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
     case SetPurgeSeqMsg(newPurgeSeq: Long) =>
       pendingPurgeSeq = newPurgeSeq
       logger.debug(prefix_name("purge sequence is now %d".format(newPurgeSeq)))
-      'ok
+      if (shouldCommit(pendingSeq, newPurgeSeq)) {
```
nickva: We may still want to tweak this a bit to get better guarantees that when we return from this call we have set the purge sequence and committed successfully that particular update. It seems currently `shouldCommit` would return false if there is another commit in progress (`committing = true`), in  that case, we'd set a pending purge sequence but not commit and if the index crashes we could regress back.

nickva (commented):
Could work but may need a few more tweaks to ensure we do not regress the purge sequence if we skipped the `shouldCommit` part in that call sequence.

Since set_purge commit will also drive commits not just periodic commits, before a periodic commit we can check if we have just committed recently (in less than 1/2 of the period or so) and skip it. In other words we could end up with setting a purge seq and force committing and them the 1 second commit time forcing another commit and perhaps if commits are a bit heavy we can delay periodic one until the next period

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
rnewson: agreed. need to only send 'ok' to a waiter if we committed their purge_seq value (or a higher one)

rnewson: ok, so I've revised this.

if you set_purge_seq to a higher value than the index has committed;

1. we add your 'from' to a list, recording which purge seq you wanted to set
2. we attempt to start the commit, which will happen as long as we're not currently waiting for a commit to complete
3. when commits complete (success or failure) we also check if we should immediately start a new commit. A commit that wasn't started at step 2 will be started in this step as the criteria is that we are not currently committed and that at least one of the 'pending' sequences has been advanced, which it was in step 1.

nickva (approved):
+1 thank you!

I added some zeunit tests for it in a pr to this pr https://github.com/cloudant-labs/clouseau/pull/150 but it might be better to just copy them in, no need to keep a separate commit just for it

PR #147, opened 2026-03-17
Fix partition timeout feature

rnewson:
We had a singleton QueryTimeoutImpl so after 5 seconds all searches timeout. This is manifested by immediately returning 0 hits(!!)

So, create a new QueryTimeoutImpl for each query and also check .timedOut and throw an error if exceeded

### DIFF
--- a/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
+++ b/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
@@ -98,15 +98,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
 
   // Start committer heartbeat
   val commitInterval = ctx.args.config.getInt("commit_interval_secs", 30)
-  val timeAllowed = new QueryTimeoutImpl(ctx.args.config.getLong("clouseau.search_allowed_timeout_msecs", 5000)) {
-    override def shouldExit(): Boolean = {
-      val result = super.shouldExit()
-      if (result) {
-        parSearchTimeOutCount += 1
-      }
-      result
-    }
-  }
+  val timeAllowed = ctx.args.config.getLong("clouseau.search_allowed_timeout_msecs", 5000)
   val countFieldsEnabled = ctx.args.config.getBoolean("clouseau.count_fields", false)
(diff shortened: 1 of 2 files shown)


nickva (approved):
Nice find! And it explains why this was a partitioned db-only problem and the 3.x code I see has a different way of solving the same problem, which explains why this is 4.x only problem.

+1 from me, but may be good to get someone else who understands Scala to take a better look

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -399,13 +390,17 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
           }
 
           if (partition.isDefined) {
-            searcher.setTimeout(timeAllowed)
+            searcher.setTimeout(timeAllowed())
```
pgj: Is this the parentheses needed?  I am not sure about that.  Internet tells me parentheses can be omitted when there are no parameters, however if the function has side effects, i.e. "it does something", parentheses shall be added.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -436,6 +431,18 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
     }
   }
 
+  private def timeAllowed(): QueryTimeout = {
```
pgj: Same here: I am not sure if parenthesis needed.  To me, it is a definition with no side effects.  It basically produces an object (a value) but that is dependent on the configuration settings.  There is no support in Clouseau to reload the configuration during run time, so that the setting involved is technically a global constant in this case.

pgj:
> +1 from me, but may be good to get someone else who understands Scala to take a better look

The Scala part does not look that problematic, albeit I added some comments.  I am more worried about the test coverage: this is a bug that slipped through the cracks of the current set of tests so it would be beneficial to pair up the change with a test somewhere (it can be in CouchDB upstream as an Elixir test) that checks for this regression specifically.

pgj (approved):

PR #149, opened 2026-03-19
clouseau: implement message forwarding for individual indexers

pgj:
Clouseau maintains an LRU cache of actively managed indexes in parallel to Dreyfus (on the CouchDB side), which caches may get out of sync if the rate of eviction is too high.

Such situations may happen when there are too many indexes to manage in a relatively short period of time so that Clouseau is forced to switch between them very rapidly and therefore it becomes a subject of thrashing.  Thrashing may lead to process termination messages of indexers being closed (by the unexpected eviction) not necessarily processed in time and the subsequent search requests on the Dreyfus side will start failing via observing `noproc` errors due to the backing indexer is gone. This immediately translates to an HTTP 500 errors, which is visible to the user.

It is possible to mitigate the problem by raising the size of the LRU cache high enough (by setting the `max_indexes_open` configuration value) so that the chance of eviction becomes low enough to let the indexer termination messages propagate timely. One may consult the recently introduced set of metrics (in #132) to understand the optimal size of the cache with regard to the actual workload.

But that cannot be considered the perfect remedy.  Maintaining a shared state joined by Erlang process identifiers as keys is an inherently fragile idea due to their volatile nature when one of the sides is emulated in Java.  That is why we shall find a more stable key, the path of the index file which is less prone to change.

Indexes would be still managed by dedicated indexers processes as before, but the communication to them would happen solely through a third-party process on the Clouseau side.  This process would then keep track of the states and availability of each indexer and handle the message forwarding for them properly.

The change is implemented in a backward compatible way by incrementally extending the existing Dreyfus-Clouseau interface so that the old-style message passing would just work with CouchDB instances where the new protocol is not yet supported or disabled.

### DIFF
--- a/clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
+++ b/clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
@@ -31,7 +31,6 @@ case class SearchRequest(options: Map[Symbol, Any])            extends ClouseauM
 case class SetPurgeSeqMsg(seq: Long)                           extends ClouseauMessage
 case class SetUpdateSeqMsg(seq: Long)                          extends ClouseauMessage
 case class UpdateDocMsg(id: String, doc: Document)             extends ClouseauMessage
-case class ForwardByPathMsg(path: String, msg: Any)            extends ClouseauMessage
 
 object ClouseauTypeFactory extends TypeFactory {
   type T = ClouseauMessage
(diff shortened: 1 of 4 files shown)


clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauTypeFactory.scala
```diff
@@ -151,6 +152,8 @@ object ClouseauTypeFactory extends TypeFactory {
         seq.toLong.map(SetPurgeSeqMsg)
       case ETuple(EAtom("set_update_seq"), seq: ENumber) =>
         seq.toLong.map(SetUpdateSeqMsg)
+      case ETuple(EAtom("forward"), path: EBinary, msg) =>
```
rnewson: Why is is necessary to add a clause here for every message elsewhere? This seems like a huge architectural mistake to me. For the other examples it is a desire to convert numbers of any magnitude to a long (which for `committed` I simply did in IndexService itself and not here). In this example you receive a binary and convert to a string, but you could do that in IndexService too, less duplicative, less magical.

pgj: It is not necessary.  I just thought it to be a good practice to translate this message to the corresponding constructor for the `ClouseauMessage` [sum type](https://en.wikipedia.org/wiki/Sum_type).  Fortunately, there is not even necessary to do anything about the `EBinary` to `String` conversion as it will be handled automatically.

PR #150, opened 2026-03-19
Add set and get purge_seq test to zeunit

nickva:
Now that we wait for the update to commit we can check that the purge_seq we get is always reflected in the one we just set.

Also, add a few cases where we check it doesn't regress and that we can use a large long value.


nickva:
This was integrated into the set_purge PR. Closing

PR #152, opened 2026-03-23
Add a log line when `close_if_idle` is enabled

jiahuili430:
To ensure consistency between Clouseau 3.x and 2.x, log `idle_check_interval_secs`
to the console when `close_if_idle` is enabled.


pgj (approved):
Looks great, thank you for spotting this!

PR #153, opened 2026-03-23
fix(`scripts`): find the Clouseau process more reliably for restart test

pgj:
In `restart-test`, the process to kill is selected via filtering the Java processes discovered by `jcmd`.  This may not always be reliable, and for example, if the tests are run in a container hosted atop Kubernetes in a pod with `clouseau` in its name, equipped with a Jenkins agent (which has the pod name as a command-line argument), the agent itself might get killed instead of Clouseau in result.

Make the discovery of the Clouseau process more robust by tightening the regular expression used for filtering so that the chance of similar mishaps becomes lower.

### DIFF
--- a/scripts/restart-test
+++ b/scripts/restart-test
@@ -31,7 +31,7 @@ for i in $(seq 1 "$RETRY"); do
       break
     fi
   done
-  jcmd | grep clouseau | cut -d' ' -f1 >"$TMP_DIR/t$i.pid"
+  pgrep -f "^java -jar (.*/)?clouseau.*\.jar$" >"$TMP_DIR/t$i.pid"
   before_kill_timestamp=$(date +%s)
   pkill -F "$TMP_DIR/t$i.pid"
 


scripts/restart-test
```diff
@@ -31,7 +31,7 @@ for i in $(seq 1 "$RETRY"); do
       break
     fi
   done
-  jcmd | grep clouseau | cut -d' ' -f1 >"$TMP_DIR/t$i.pid"
+  jcmd | grep -E "^[0-9]* (.*/)?clouseau.*\.jar$" | cut -d' ' -f1 >"$TMP_DIR/t$i.pid"
   before_kill_timestamp=$(date +%s)
   pkill -F "$TMP_DIR/t$i.pid"
```
rnewson, line 36: iirc pkill can take a lot of arguments to fine-tune which processes it matches, not least the one that matches against the process name. Could this section be simpler and safer if it did that instead?

pgj: The PID file is used in the later sections of the script, so I am inclined to believe that the invocation of `jcmd` and the subsequent instance of `pkill` cannot be just merged.  Of course, perhaps `pgrep` might be possible to used for this purpose in that case still.

jiahuili430 (approved):

scripts/restart-test
rnewson: nice

PR #155, opened 2026-03-25
Cleanup Configuration

jiahuili430:
- Remove `_root_` from import statements
- Remove `final` from case objects
- Remove `asInstanceOf[String]`
- Remove unused metric `field_cache_metrics`
- Reorder `WorkerConfiguration`
- Refactor `CapacityConfiguration` and extract some comments
- Update tests in `ConfigSpec.scala`

### DIFF
--- a/clouseau/src/test/scala/com/cloudant/ziose/clouseau/ConfigSpec.scala
+++ b/clouseau/src/test/scala/com/cloudant/ziose/clouseau/ConfigSpec.scala
@@ -3,18 +3,17 @@ sbt 'clouseau/testOnly com.cloudant.ziose.clouseau.ConfigSpec'
  */
 package com.cloudant.ziose.clouseau
 
-import org.junit.runner.RunWith
-import zio._
-import zio.test.junit.{JUnitRunnableSpec, ZTestJUnitRunner}
-
-import zio.test._
-import zio.test.Assertion._
 import com.cloudant.ziose.core.Exponent
 import com.cloudant.ziose.test.helpers.TestRunner
+import org.junit.runner.RunWith
+import zio.test.Assertion.{equalTo, isSome}
+import zio.test.{assert, assertTrue}
+import zio.test.junit.{JUnitRunnableSpec, ZTestJUnitRunner}
+import zio.{Chunk, Config, LogLevel}
 
 @RunWith(classOf[ZTestJUnitRunner])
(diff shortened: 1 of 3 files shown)


clouseau/src/test/scala/com/cloudant/ziose/clouseau/ConfigSpec.scala
```diff
@@ -27,87 +25,55 @@ class ConfigSpec extends JUnitRunnableSpec {
...
     }
 
     val validTests = (1 to 16).map(idx => {
       test(s"Ensure we can get correct exponent - ${idx}")(
         for {
           config <- AppCfg.fromHoconString(capacityFixture(key, idx))
-        } yield assert(getCapacity(config))(
-          isSome(equalTo(Exponent(idx)))
+        } yield assertTrue(
+          getCapacity(config).get == Exponent(idx)
```
pgj: Will not it cause a run-time error (exception raised) if the returned value here is `None`?  I think `isSome` could help with gracefully fail the test case even in those scenarios too.

jiahuili430: Good point! `isSome()` is able to handle `None`, but if we do not specify `capacity` in the config file, this test still fail, because `None` is not equal to `Exponent(idx)`.
- Test branch: https://github.com/cloudant-labs/clouseau/compare/main...jiahuili430:clouseau:t-code-cleanup
- CI: https://github.com/jiahuili430/clouseau/actions/runs/24115379653

Instead, I added [a test case for when the capacity is not specified](https://github.com/cloudant-labs/clouseau/pull/155/changes#diff-2ce88162c2fefaf957d3a1bc867f34d83a81f59881a76c02150f130c19cb8763R71).
Thanks for the review!

pgj: Yeah, that is a good idea.  But I still think we should leave this part untouched.  It is not nice to get a (potentially) lengthy stack trace with an exception in case of a failure.  Let us stick to the Boolean verdict, and land with `false` for cases where the result is `None` (possibly by accident).

jiahuili430: Updated, thanks for the review.

pgj (approved):
+1 if passes CI

PR #156, opened 2026-04-01
Use ZioSupport syntactic sugar

jiahuili430:
**Use ZioSupport syntactic sugar:**
- Improve `ZioSupport` and migrate it to the core package.
- Replace `unsafe.run()` with the corresponding function in `ZioSupport`.
- Update types and import statements


clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
```diff
@@ -50,68 +42,57 @@ class ClouseauNode(implicit
...
-          } yield addressable
-        )
-        .getOrThrowFiberFailure
-    }
-    result.self
+    val addressableActor: AddressableActor[_ <: Actor, _ <: ProcessContext] = (
+      for {
+        addressable <- worker.spawn(SimpleProcess.make(this, fun))
+      } yield addressable
+    ).unsafeRunCustomRuntimeGetOrThrowFiberFailure(runtime)
```
pgj: I am wondering if it would be sufficient to only wrap the `unsafeRun` call in `unsafeRunCustomRuntime` and let `getOrThrowFiberFailure` to be attached separately.  It maintains compositionality better:

```scala
val addressableActor: AddressableActor[_ <: Actor, _ <: ProcessContext] = (
  for {
    addressable <- worker.spawn(SimpleProcess.make(this, fun))
  } yield addressable
).unsafeRunWith(runtime).getOrThrowFiberFailure
```

I would also shorten `unsafeRunCustomRuntime` to `unsafeRunWith` since that goes better with the parameter (see above).

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
```diff
@@ -50,68 +42,57 @@ class ClouseauNode(implicit
...
+    addressableActor.self
   }
 
-  override def spawnService[TS <: Service[A] with Actor: Tag, A <: Product](
+  override def spawnService[TS <: Service[A] & Actor: Tag, A <: Product](
```
pgj, line 53: Is the use `&` common in other parts of the code?  I am worried that we are mixing the two syntax over the source code.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
```diff
@@ -50,68 +42,57 @@ class ClouseauNode(implicit
...
-    } // TODO: kill the caller
+  )(implicit adapter: Adapter[_, _]): Result[Node.Error, AddressableActor[TS, ProcessContext]] = {
+    val result = spawnServiceZIO[TS, A](builder).unsafeRunCustomRuntime(runtime) // TODO: kill the caller
     result match {
       case Failure(cause) if cause.isFailure     => core.Failure(cause.failureOption.get)
-      case Failure(cause) if cause.isDie         => core.Failure(core.Node.Error.Unknown(cause.dieOption.get))
-      case Failure(cause) if cause.isInterrupted => core.Failure(core.Node.Error.Interrupt(cause.interruptOption.get))
-      case Failure(cause: core.Node.Error)       => core.Failure(cause)
-      case Failure(cause) => core.Failure(core.Node.Error.Unknown(new Throwable(cause.prettyPrint)))
+      case Failure(cause) if cause.isDie         => core.Failure(Node.Error.Unknown(cause.dieOption.get))
```
pgj, line 59: Why did not you remove the `core.` qualifier from `Failure`?

clouseau/src/main/scala/com/cloudant/ziose/clouseau/LoggerFactory.scala
```diff
@@ -115,8 +117,8 @@ object LoggerFactory {
       }
     }
 
-    def log(event: UIO[Unit])(implicit adapter: Adapter[_, _]): Unit = {
-      Unsafe.unsafe(implicit u => adapter.runtime.unsafe.run(event.timeout(Duration.fromSeconds(10)).forkDaemon))
+    def log(event: UIO[Unit])(implicit adapter: Adapter[_, _]): UIO[Fiber.Runtime[Nothing, Option[Unit]]] = {
+      event.timeout(Duration.fromSeconds(10)).unsafeRunAdapterForkDaemon
```
pgj: I would advise against defining `unsafeRunAdapterForkDaemon` — cannot we just express the same with `unsafeRunWith(adapter.runtime).forkDaemon`?

core/src/main/scala/com/cloudant/ziose/core/EngineWorker.scala
```diff
@@ -29,23 +29,15 @@ trait EngineWorker extends ForwardWithId[Engine.WorkerId, MessageEnvelope] {
   }
 
   def processInfo[A <: Actor](addr: Address): Option[ProcessInfo] = {
-    Unsafe.unsafe { implicit unsafe =>
-      Runtime.default.unsafe
-        .run(processInfoZIO(addr))
-        .getOrThrowFiberFailure()
-    }.asInstanceOf[Option[ProcessInfo]]
+    processInfoZIO(addr).unsafeRunGetOrThrowFiberFailure
```
pgj: How about writing `unsafeRunWith(Runtime.default).getOrThrowFiberFailure` instead?

core/src/main/scala/com/cloudant/ziose/core/EngineWorker.scala
```diff
@@ -29,23 +29,15 @@ trait EngineWorker extends ForwardWithId[Engine.WorkerId, MessageEnvelope] {
...
     exchange.actorMeters(addr)
   }
 
   def actorMeters[A <: Actor](addr: Address): Option[List[ActorMeterInfo]] = {
-    Unsafe.unsafe { implicit unsafe =>
-      Runtime.default.unsafe
-        .run(actorMetersZIO(addr))
-        .getOrThrowFiberFailure()
-    }.asInstanceOf[Option[List[ActorMeterInfo]]]
+    actorMetersZIO(addr).unsafeRunGetOrThrowFiberFailure
```
pgj: Please consider writing `unsafeRunWith(Runtime.default).getOrThrowFiberFailure` here.

core/src/main/scala/com/cloudant/ziose/core/EngineWorker.scala
```diff
@@ -63,11 +55,7 @@ trait EngineWorker extends ForwardWithId[Engine.WorkerId, MessageEnvelope] {
   }
 
   def processInfoTopK(valueFun: ProcessInfo => Int): List[ProcessInfo] = {
-    Unsafe.unsafe { implicit unsafe =>
-      Runtime.default.unsafe
-        .run(processInfoTopKZIO(valueFun))
-        .getOrThrowFiberFailure()
-    }.asInstanceOf[List[ProcessInfo]]
+    processInfoTopKZIO(valueFun).unsafeRunGetOrThrowFiberFailure
```
pgj: Please consider writing `unsafeRunWith(Runtime.default).getOrThrowFiberFailure` here.

core/src/main/scala/com/cloudant/ziose/core/EngineWorker.scala
```diff
@@ -87,13 +75,8 @@ trait EngineWorker extends ForwardWithId[Engine.WorkerId, MessageEnvelope] {
   }
 
   def actorMeterInfoTopK(query: ActorMeterInfo.Query[Double]): List[ActorMeterInfo] = {
-    Unsafe.unsafe { implicit unsafe =>
-      Runtime.default.unsafe
-        .run(actorMeterInfoTopKZIO(query))
-        .getOrThrowFiberFailure()
-    }.asInstanceOf[List[ActorMeterInfo]]
+    actorMeterInfoTopKZIO(query).unsafeRunGetOrThrowFiberFailure
```
pgj: Please consider writing `unsafeRunWith(Runtime.default).getOrThrowFiberFailure` here.

core/src/main/scala/com/cloudant/ziose/core/ZioSupport.scala
```diff
@@ -1,9 +1,47 @@
...
 
 trait ZioSupport {
-  implicit final class ZioOps[E, A](self: IO[E, A]) {
-    def unsafeRun: A = Unsafe.unsafe(implicit u => Runtime.default.unsafe.run(self).getOrThrowFiberFailure())
+  implicit final class ZioOps[E <: Throwable, A](self: IO[E, A]) {
+    def unsafeRun: Exit[E, A] = {
+      Unsafe.unsafe(implicit u => Runtime.default.unsafe.run(self))
+    }
+
+    def unsafeRunGetOrThrowFiberFailure: A = {
```
pgj: There is no need for this definition.  This is not compositional: one could easily express same simply by attaching the `.getOrThrowFiberFailure` to the invocation of the `unsafeRun` function locally ("ad hoc").  One does not want to introduce a separate wrapper for every combination.

core/src/main/scala/com/cloudant/ziose/core/ZioSupport.scala
```diff
@@ -1,9 +1,47 @@
...
+      Unsafe.unsafe(implicit u => Runtime.default.unsafe.run(self).getOrThrowFiberFailure())
+    }
+  }
+
+  implicit final class ZioOpsAdapter[E <: Throwable, A](self: IO[E, A]) {
```
pgj: I would lose this whole class.  Based on what I wrote above, there is no need for this one either.

core/src/main/scala/com/cloudant/ziose/core/ZioSupport.scala
```diff
@@ -1,9 +1,47 @@
...
+      Unsafe.unsafe(implicit u => adapter.runtime.unsafe.run(self).forkDaemon)
+    }
+
+    def unsafeRunAdapterGetOrThrow(implicit adapter: Adapter[_, _]): A = {
+      Unsafe.unsafe(implicit u => adapter.runtime.unsafe.run(self).getOrThrow())
+    }
+  }
+
+  implicit final class ZioOpsCustomRuntime[R, E <: Throwable, A](self: ZIO[R, E, A]) {
+    def unsafeRunCustomRuntime(runtime: Runtime[R]): Exit[E, A] = {
```
pgj: Please consider renaming this to `unsafeRunWith` and lose the other two definitions for the better user experience.

scalang/src/main/scala/com/cloudant/ziose/scalang/Adapter.scala
```diff
@@ -13,7 +13,7 @@ class Adapter[C <: ProcessContext, F <: TypeFactory] private (
   type_factory: Option[F],
   val workerId: Engine.WorkerId,
   val workerNodeName: Symbol
-) {
+) extends core.Adapter[C, F] {
```
pgj: Why do we need `Adapter[_, _]` here?  Can we lose the `core.` prefix?

scalang/src/main/scala/com/cloudant/ziose/scalang/MetricsGroup.scala
```diff
@@ -26,9 +26,9 @@ case class MetricsGroup(klass: Class[_], metricsRegistry: ScalangMeterRegistry)
       for {
         _ <- zioCounter.update(delta)
       } yield ()
-    ).unsafeRun
-    def -=(delta: Int): Unit       = zioCounter.update(-delta.abs).unsafeRun
-    def count: MetricState.Counter = zioCounter.value.unsafeRun
+    ).unsafeRunGetOrThrowFiberFailure
```
pgj: Please consider writing `unsafeRun.getOrThrowFiberFailure` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -181,43 +159,35 @@ trait ProcessLike[A <: Adapter[_, _]] extends core.Actor {
     }
   }
 
-  def unlink(to: Pid): Unit = {
-    Unsafe.unsafe { implicit unsafe =>
-      adapter.runtime.unsafe.run(adapter.unlink(to.fromScala))
-    }
+  def unlink(to: Pid): UIO[Unit] = {
+    adapter.unlink(to.fromScala).unsafeRunAdapter(adapter)
```
pgj: Please consider writing `.unsafeRunWith(adapter.runtime)` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -181,43 +159,35 @@ trait ProcessLike[A <: Adapter[_, _]] extends core.Actor {
...
   }
 
   def link(to: Pid): Unit = {
-    Unsafe.unsafe { implicit unsafe =>
-      adapter.runtime.unsafe.run(adapter.link(to.fromScala)).getOrThrow()
-    }
+    adapter.link(to.fromScala).unsafeRunAdapterGetOrThrow(adapter)
```
pgj: Please consider writing `.unsafeRunWith(adapter.runtime).getOrThrow` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -181,43 +159,35 @@ trait ProcessLike[A <: Adapter[_, _]] extends core.Actor {
...
     } yield Reference.toScala(ref)
   }
 
   def monitor(monitored: Any): Reference = {
-    Unsafe.unsafe { implicit unsafe =>
-      val rt = adapter.runtime.asInstanceOf[Runtime[core.Node]]
-      rt.unsafe.run[Node.Error, Reference](monitorZIO(monitored)).getOrThrow()
-    }
+    val rt: Runtime[Node] = adapter.runtime.asInstanceOf[Runtime[Node]]
+    monitorZIO(monitored).unsafeRunCustomRuntimeGetOrThrow(rt)
```
pgj: Please consider writing `.unsafeRunWith(rt).getOrThrow` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -181,43 +159,35 @@ trait ProcessLike[A <: Adapter[_, _]] extends core.Actor {
...
   }
 
-  def demonitor(ref: Reference) = {
-    Unsafe.unsafe { implicit unsafe =>
-      adapter.runtime.unsafe.run(adapter.demonitor(ref.fromScala))
-    }
+  def demonitor(ref: Reference): UIO[Unit] = {
+    adapter.demonitor(ref.fromScala).unsafeRunAdapter(adapter)
```
pgj: Please consider writing `.unsafeRunWith(adapter.runtime)` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -270,17 +240,19 @@ class Process(implicit val adapter: Adapter[_, _]) extends ProcessLike[Adapter[_
 
   def sendEvery(pid: Pid, msg: Any, delay: Long) = {
     val interval: Duration = Duration.fromMillis(delay)
-    adapter.forkScoped(sendZIO(pid, msg).schedule(Schedule.spaced(interval))).unsafeRun
+    adapter.forkScoped(sendZIO(pid, msg).schedule(Schedule.spaced(interval))).unsafeRunGetOrThrowFiberFailure
```
pgj: Please consider writing `.unsafeRun.getOrThrowFiberFailure` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -270,17 +240,19 @@ class Process(implicit val adapter: Adapter[_, _]) extends ProcessLike[Adapter[_
...
   }
 
   def sendEvery(name: Symbol, msg: Any, delay: Long) = {
     val interval: Duration = Duration.fromMillis(delay)
-    adapter.forkScoped(sendZIO(name, msg).schedule(Schedule.spaced(interval))).unsafeRun
+    adapter.forkScoped(sendZIO(name, msg).schedule(Schedule.spaced(interval))).unsafeRunGetOrThrowFiberFailure
```
pgj: Please consider writing `.unsafeRun.getOrThrowFiberFailure` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -270,17 +240,19 @@ class Process(implicit val adapter: Adapter[_, _]) extends ProcessLike[Adapter[_
...
   }
 
   def sendEvery(dest: (RegName, NodeName), msg: Any, delay: Long) = {
     val interval: Duration = Duration.fromMillis(delay)
-    adapter.forkScoped(sendZIO(dest, Pid.toScala(self.pid), msg).schedule(Schedule.spaced(interval))).unsafeRun
+    adapter
+      .forkScoped(sendZIO(dest, Pid.toScala(self.pid), msg).schedule(Schedule.spaced(interval)))
+      .unsafeRunGetOrThrowFiberFailure
```
pgj: Please consider writing `unsafeRun.getOrThrowFiberFailure` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -617,17 +590,17 @@ object Service {
     }
   }
 
-  def call(to: core.Address, msg: Any)(implicit adapter: Adapter[_, _]): Any = {
-    val result = Unsafe.unsafe { implicit unsafe =>
-      val rt = adapter.runtime.asInstanceOf[Runtime[core.Node]]
-      rt.unsafe.run(callZIO(to, msg))
+  def call(to: Address, msg: Any)(implicit adapter: Adapter[_, _]): Any = {
+    val rt: Runtime[Node] = adapter.runtime.asInstanceOf[Runtime[Node]]
```
pgj: I wonder why `adapter.runtime` is not of the right type, and why this casting is needed.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -617,17 +590,17 @@ object Service {
...
     replyFromCall(result)
   }
-  def call(to: core.Address, msg: Any, timeout: Long)(implicit adapter: Adapter[_, _]): Any = {
-    val result = Unsafe.unsafe { implicit unsafe =>
-      val rt = adapter.runtime.asInstanceOf[Runtime[core.Node]]
-      rt.unsafe.run(callZIO(to, msg, Some(Duration.fromMillis(timeout))))
+  def call(to: Address, msg: Any, timeout: Long)(implicit adapter: Adapter[_, _]): Any = {
+    val rt: Runtime[Node] = adapter.runtime.asInstanceOf[Runtime[Node]]
+    val result: Exit[Node.Error, MessageEnvelope.Response] = {
+      callZIO(to, msg, Some(Duration.fromMillis(timeout))).unsafeRunCustomRuntime(rt)
```
pgj: Please consider writing `unsafeRunWith(rt)` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -664,27 +637,25 @@ object Service {
...
   }
 
-  def cast(to: core.Address, msg: Any)(implicit adapter: Adapter[_, _]): Unit = {
-    val result = Unsafe.unsafe { implicit unsafe =>
-      val rt = adapter.runtime.asInstanceOf[Runtime[core.Node]]
-      rt.unsafe.run(castZIO(to, msg))
-    }
+  def cast(to: Address, msg: Any)(implicit adapter: Adapter[_, _]): Exit[Node.Error, Unit] = {
+    val rt = adapter.runtime.asInstanceOf[Runtime[Node]]
+    castZIO(to, msg).unsafeRunCustomRuntime(rt)
```
pgj: Please consider writing `.unsafeRunWith(rt)` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/SNode.scala
```diff
@@ -59,51 +70,46 @@ class SNode(val metricsRegistry: ScalangMeterRegistry, val logLevel: LogLevel)(i
       case (false, false) => return false
     }
 
-    Unsafe.unsafe { implicit unsafe =>
-      adapter.runtime.unsafe.run(adapter.link(msg))
-    }
-    return true
+    adapter.link(msg).unsafeRunAdapter(adapter)
```
pgj: Please consider writing `.unsafeRunWith(adapter.runtime)` instead.

scalang/src/main/scala/com/cloudant/ziose/scalang/SNode.scala
```diff
@@ -59,51 +70,46 @@ class SNode(val metricsRegistry: ScalangMeterRegistry, val logLevel: LogLevel)(i
...
+              terminator(address, process.adapter) *> process.adapter
+                .lookUpName(name)
+                .delay(100.milliseconds)
+                .map(!_.contains(address))
+                .repeatUntil(_ == true)
+                .timeout(3.seconds)
+            case None => ZIO.some(false)
+          }
+          _ <- resultChannel.offer(res.getOrElse(false))
+        } yield ()).unsafeRunCustomRuntime(runtime)
```
pgj: Please consider writing `.unsafeRunWith(runtime)` instead.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
jiahuili430: Great idea! After using `unsafeRunWith()`, the code looks much cleaner! https://github.com/cloudant-labs/clouseau/pull/156/changes/e69000ad7e671e6682bd296dad02c50e2fe78326.

After extracting `getOrThrowFiberFailure()`, it requires an implicit `Unsafe` parameter.
However, the scope of `Unsafe` appears to have changed. Although the tests have passed,
I am not certain whether this will lead to a change in behavior.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
jiahuili430: Yes, we used `&` in different packages, just keep them consistent.

scalang/src/main/scala/com/cloudant/ziose/scalang/Adapter.scala
jiahuili430: ZioSupport resides in the `core` package, while Adapter is located in the `scalang` package.
Since `scalang` depends on `core`, I need to define the Adapter trait in the `core` package
to avoid creating a circular dependency.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
jiahuili430: The type of `adapter.runtime` is `Any`, but the compiler requires `Runtime[Node]` here.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
jiahuili430: [Replace other functions with unsafeRunWith()](https://github.com/cloudant-labs/clouseau/pull/156/changes/e69000ad7e671e6682bd296dad02c50e2fe78326) and extract `getOrThrowFiberFailure()` can pass tests, see CI: https://github.com/jiahuili430/clouseau/actions.

However, [extracting `getOrThrow()` from `unsafeRunWith()`](https://github.com/cloudant-labs/clouseau/pull/156/changes/10637fe934816092466a404c56a3b33723e0c6ef) causes the tests to fail, as it alters the types in `Service.scala`. Should I remove this commit?

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
jiahuili430: ZIO has its own `Failure` type; removing `core.` would result in a conflict.

pgj: You can always hide the conflicting symbol if it is not used in the same module.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -305,23 +281,22 @@ class Process(implicit val adapter: Adapter[_, _]) extends ProcessLike[Adapter[_
...
     event.getPayload match {
       case None        => ZIO.succeed(handleMessage(())).as(ActorResult.Continue())
       case Some(value) => ZIO.succeed(handleMessage(adapter.toScala(value))).as(ActorResult.Continue())
     }
   }
 
   def onTermination[PContext <: ProcessContext](reason: ETerm, ctx: PContext): UIO[Unit] = {
-    ZIO.logError(s"${getClass.toString} did not define a onTermination function.")
-    ZIO.succeed(())
+    ZIO.logError(s"${getClass.toString} did not define a onTermination function.").unsafeRun
```
pgj: Why is `.unsafeRun` needed here?  That is a ZIO effect.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -721,9 +695,7 @@ object Service {
     adapter.send(envelope)
   }
 
-  def reply[P <: Process](caller: (Pid, Any), reply: Any)(implicit process: P): Unit = {
-    Unsafe.unsafe { implicit unsafe =>
-      Runtime.default.unsafe.run(replyZIO(caller, reply))
-    }
+  def reply[P <: Process](caller: (Pid, Any), reply: Any)(implicit process: P): UIO[Unit] = {
```
pgj: The return type has changed to `UIO[Unit]`, which marks a ZIO effect, while `.unsafeRun` is used in the implementation.  Why?

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
jiahuili430: Without `unsafeRun`, probably we'll need to `println()` here to invoke the logger.

Maybe I'm wrong, but based on my understanding, this function is intended to be overridden by subclasses. If the subclass doesn't implement it, we will fall back and print an error message indicating that this specific sub-service didn't define the `onTermination` function.

Do we want to print out that message? Maybe I should change it to `ZIO.logError("...") *> ZIO.unit `?

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
jiahuili430: Good catch, I'll change it back.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
pgj: I am not sure I follow your argument about `println`.  Logging should just work without printing anything since the output might be something else (e.g. Syslog).  `ZIO.logError` is an effect that has the `ZIO[_, _, _]` type and it should only be "run" (via `.unsafeRun`) if the call for that is injected into an environment where ZIO effects are not used (therefore the types would not match).

The return type of `onTermination` is `UIO[Unit]` which is an alias for `ZIO[Any, Nothing, Unit]` which suggests that the expression in the body should be of this type.  I am inclined to believe that even though when `.unsafeRun` is used the type might still fit, that is an unnecessary step and it may throw an exception.

Note that [type of `ZIO.logError` already fits well `UIO[Unit]`](https://zio.dev/api/zio/zio$#logError(message:=%3EString)(implicittrace:zio.Trace):zio.UIO[Unit]) so `*> ZIO.unit` would not even needed.

That said, I think the best you could do here is just removing the line with `ZIO.succeed(())`.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
jiahuili430: The first `Failure` comes from `zio.Exit.Failure`, while the second comes from `core`.
If remove `core`, we would need to modify this line to
```scala
case zio.Exit.Failure(cause) if cause.isFailure => Failure(cause.failureOption.get)
```
I think keeping it as is might be clearer.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
jiahuili430: The tests failed in CI, so decided to continue using `unsafeRunCustomRuntimeGetOrThrow()`.

scalang/src/main/scala/com/cloudant/ziose/scalang/MetricsGroup.scala
```diff
@@ -47,7 +48,7 @@ case class MetricsGroup(klass: Class[_], metricsRegistry: ScalangMeterRegistry)
   def counter(name: String): Counter = {
     val zioCounter = Metric.counter(constructName(name))
     // register to JMX
-    zioCounter.update(0).unsafeRun
+    zioCounter.update(0).unsafeRun.getOrThrowFiberFailure()(Unsafe)
```
pgj: Why did you need to extend the invocation of `.unsafeRun` with `.getOrThrowFiberFailure`?

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -181,43 +159,35 @@ trait ProcessLike[A <: Adapter[_, _]] extends core.Actor {
...
-    }
+  def unlink(to: Pid): UIO[Unit] = {
+    adapter.unlink(to.fromScala).unsafeRunWith(adapter.runtime)
   }
 
   def link(to: Pid): Unit = {
-    Unsafe.unsafe { implicit unsafe =>
-      adapter.runtime.unsafe.run(adapter.link(to.fromScala)).getOrThrow()
-    }
+    adapter.link(to.fromScala).unsafeRunCustomRuntimeGetOrThrow(adapter.runtime)
```
pgj: Do we need to deal with calling `getOrThrow` here?  Would not `getOrThrowFiberFailure` be sufficient here?  All the relevant Java exceptions are translated to `Node.Error`s which makes me believe that actually no `Throwable` would travel here...?

pgj:
All right, I spent some time on thinking through how the various variations on `ZioSupport#unsafeRun` should be implemented.  I can acknowledge that my suggestion about maintaining the compositionality is not easy to implement and it would not give us much better user experience.  At the same time, I realized that we have basically 3 use cases that would have to be covered:

- `unsafeRun`: run the effect for itself and ignore the return value, e.g. print something to the screen.
- `unsafeRunAndGet`: run the effect for obtaining some return value, e.g. read the value of a JMX counter.
- `unsafeRunAndGetOrThrow`: similar to `unsafeRunAndGet`, i.e. run the effect for obtaining some return value but forward (re-throw) any exception raised as part of that when integrating with non-ZIO code, , e.g. a reference after issuing a monitor request.

With those cases identified, it becomes possible to define the `*With` family of extension where the custom runtime could be passed on: `unsafeRunWith`, `unsafeRunAndGetWith`, and `unsafeRunAndGetOrThrowWith`.  This way the call site becomes clearer and simpler.

Based on that, my suggestion formulated as a patch atop the current version of the branch is as follows:

```diff
diff --git a/clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala b/clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
index 9a5cb7d7..3561998e 100644
--- a/clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
+++ b/clouseau/src/main/scala/com/cloudant/ziose/clouseau/ClouseauNode.scala
@@ -5,7 +5,7 @@ import com.cloudant.ziose.{core, scalang}
 import core.{Actor, ActorBuilder, AddressableActor, EngineWorker, Node, ProcessContext, Result}
 import scalang.{Adapter, SNode, ScalangMeterRegistry, Service}
 import zio.Exit.{Failure, Success}
-import zio.{&, LogLevel, Runtime, Tag, Unsafe, ZIO}
+import zio.{&, LogLevel, Runtime, Tag, ZIO}

 class ClouseauNode(implicit
   override val runtime: Runtime[EngineWorker & Node],
@@ -46,7 +46,7 @@ class ClouseauNode(implicit
       for {
         addressable <- worker.spawn(SimpleProcess.make(this, fun))
       } yield addressable
-    ).unsafeRunWith(runtime).getOrThrowFiberFailure()(Unsafe)
+    ).unsafeRunAndGetWith(runtime)
     addressableActor.self
   }

diff --git a/core/src/main/scala/com/cloudant/ziose/core/EngineWorker.scala b/core/src/main/scala/com/cloudant/ziose/core/EngineWorker.scala
index 3dad4207..405578d7 100644
--- a/core/src/main/scala/com/cloudant/ziose/core/EngineWorker.scala
+++ b/core/src/main/scala/com/cloudant/ziose/core/EngineWorker.scala
@@ -29,7 +29,7 @@ trait EngineWorker extends ForwardWithId[Engine.WorkerId, MessageEnvelope] with
   }

   def processInfo[A <: Actor](addr: Address): Option[ProcessInfo] = {
-    processInfoZIO(addr).unsafeRun.getOrThrowFiberFailure()(Unsafe)
+    processInfoZIO(addr).unsafeRunAndGet
   }

   def actorMetersZIO[A <: Actor](addr: Address): UIO[Option[List[ActorMeterInfo]]] = {
@@ -37,7 +37,7 @@ trait EngineWorker extends ForwardWithId[Engine.WorkerId, MessageEnvelope] with
   }

   def actorMeters[A <: Actor](addr: Address): Option[List[ActorMeterInfo]] = {
-    actorMetersZIO(addr).unsafeRun.getOrThrowFiberFailure()(Unsafe)
+    actorMetersZIO(addr).unsafeRunAndGet
   }

   def processInfoTopKZIO[A <: Actor](
@@ -55,7 +55,7 @@ trait EngineWorker extends ForwardWithId[Engine.WorkerId, MessageEnvelope] with
   }

   def processInfoTopK(valueFun: ProcessInfo => Int): List[ProcessInfo] = {
-    processInfoTopKZIO(valueFun).unsafeRun.getOrThrowFiberFailure()(Unsafe)
+    processInfoTopKZIO(valueFun).unsafeRunAndGet
   }

   def actorMeterInfoTopKZIO[A <: Actor](
@@ -75,7 +75,7 @@ trait EngineWorker extends ForwardWithId[Engine.WorkerId, MessageEnvelope] with
   }

   def actorMeterInfoTopK(query: ActorMeterInfo.Query[Double]): List[ActorMeterInfo] = {
-    actorMeterInfoTopKZIO(query).unsafeRun.getOrThrowFiberFailure()(Unsafe)
+    actorMeterInfoTopKZIO(query).unsafeRunAndGet
   }
 }

diff --git a/core/src/main/scala/com/cloudant/ziose/core/ZioSupport.scala b/core/src/main/scala/com/cloudant/ziose/core/ZioSupport.scala
index a291788c..efd0d82f 100644
--- a/core/src/main/scala/com/cloudant/ziose/core/ZioSupport.scala
+++ b/core/src/main/scala/com/cloudant/ziose/core/ZioSupport.scala
@@ -1,15 +1,19 @@
 package com.cloudant.ziose.core

-import zio.{Exit, IO, Runtime, Unsafe, ZIO}
+import zio.{Exit, Runtime, Unsafe, ZIO}

 trait Adapter[C, F] {
   def runtime: Runtime[Any]
 }

 trait ZioSupport {
-  implicit final class ZioOps[E <: Throwable, A](self: IO[E, A]) {
+  implicit final class ZioOps[E <: Throwable, A](self: ZIO[Any, E, A]) {
     def unsafeRun: Exit[E, A] = {
-      Unsafe.unsafe(implicit u => Runtime.default.unsafe.run(self))
+      self.unsafeRunWith(Runtime.default)
+    }
+
+    def unsafeRunAndGet: A = {
+      self.unsafeRunAndGetWith(Runtime.default)
     }
   }

@@ -18,8 +22,12 @@ trait ZioSupport {
       Unsafe.unsafe(implicit u => runtime.unsafe.run(self))
     }

-    def unsafeRunCustomRuntimeGetOrThrow(runtime: Runtime[R]): A = {
+    def unsafeRunAndGetOrThrowWith(runtime: Runtime[R]): A = {
       Unsafe.unsafe(implicit u => runtime.unsafe.run(self).getOrThrow())
     }
+
+    def unsafeRunAndGetWith(runtime: Runtime[R]): A = {
+      Unsafe.unsafe(implicit u => runtime.unsafe.run(self).getOrThrowFiberFailure())
+    }
   }
 }
diff --git a/scalang/src/main/scala/com/cloudant/ziose/scalang/MetricsGroup.scala b/scalang/src/main/scala/com/cloudant/ziose/scalang/MetricsGroup.scala
index e5e74f7d..d6277084 100644
--- a/scalang/src/main/scala/com/cloudant/ziose/scalang/MetricsGroup.scala
+++ b/scalang/src/main/scala/com/cloudant/ziose/scalang/MetricsGroup.scala
@@ -28,8 +28,8 @@ case class MetricsGroup(klass: Class[_], metricsRegistry: ScalangMeterRegistry)
         _ <- zioCounter.update(delta)
       } yield ()
     ).unsafeRun.getOrThrowFiberFailure()(Unsafe)
-    def -=(delta: Int): Unit       = zioCounter.update(-delta.abs).unsafeRun.getOrThrowFiberFailure()(Unsafe)
-    def count: MetricState.Counter = zioCounter.value.unsafeRun.getOrThrowFiberFailure()(Unsafe)
+    def -=(delta: Int): Unit       = zioCounter.update(-delta.abs).unsafeRun
+    def count: MetricState.Counter = zioCounter.value.unsafeRunAndGet
     def clear                      = zioCounter.fromConst(0)

     def get: Metric.Counter[Long] = this.zioCounter
@@ -48,7 +48,7 @@ case class MetricsGroup(klass: Class[_], metricsRegistry: ScalangMeterRegistry)
   def counter(name: String): Counter = {
     val zioCounter = Metric.counter(constructName(name))
     // register to JMX
-    zioCounter.update(0).unsafeRun.getOrThrowFiberFailure()(Unsafe)
+    zioCounter.update(0).unsafeRun
     new Counter(zioCounter)
   }

diff --git a/scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala b/scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
index 1781bc14..06629167 100644
--- a/scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
+++ b/scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
@@ -98,15 +98,15 @@ trait ProcessLike[A <: Adapter[_, _]] extends Actor with ZioSupport {
     Address.fromRemoteName(EAtom(name), EAtom(node), self.workerId, self.workerNodeName)
   }

-  def send(pid: Pid, msg: Any): UIO[Unit] = {
+  def send(pid: Pid, msg: Any): Unit = {
     sendZIO(pid, msg).unsafeRun
   }

-  def send(name: RegName, msg: Any): UIO[Unit] = {
+  def send(name: RegName, msg: Any): Unit = {
     sendZIO(name, msg).unsafeRun
   }

-  def send(dest: (RegName, NodeName), from: Pid, msg: Any): UIO[Unit] = {
+  def send(dest: (RegName, NodeName), from: Pid, msg: Any): Unit = {
     sendZIO(dest, from, msg).unsafeRun
   }

@@ -123,11 +123,11 @@ trait ProcessLike[A <: Adapter[_, _]] extends Actor with ZioSupport {
     adapter.send(envelope)
   }

-  def exit(pid: Pid, reason: Any): UIO[Unit] = {
+  def exit(pid: Pid, reason: Any): Unit = {
     exitZIO(pid, reason).unsafeRun
   }

-  def exit(name: RegName, reason: Any): UIO[Unit] = {
+  def exit(name: RegName, reason: Any): Unit = {
     exitZIO(name, reason).unsafeRun
   }

@@ -159,12 +159,12 @@ trait ProcessLike[A <: Adapter[_, _]] extends Actor with ZioSupport {
     }
   }

-  def unlink(to: Pid): UIO[Unit] = {
+  def unlink(to: Pid): Unit = {
     adapter.unlink(to.fromScala).unsafeRunWith(adapter.runtime)
   }

   def link(to: Pid): Unit = {
-    adapter.link(to.fromScala).unsafeRunCustomRuntimeGetOrThrow(adapter.runtime)
+    adapter.link(to.fromScala).unsafeRunAndGetOrThrowWith(adapter.runtime)
   }

   def monitorZIO(monitored: Any): ZIO[Node, _ <: Node.Error, Reference] = {
@@ -183,10 +183,10 @@ trait ProcessLike[A <: Adapter[_, _]] extends Actor with ZioSupport {

   def monitor(monitored: Any): Reference = {
     val rt: Runtime[Node] = adapter.runtime.asInstanceOf[Runtime[Node]]
-    monitorZIO(monitored).unsafeRunCustomRuntimeGetOrThrow(rt)
+    monitorZIO(monitored).unsafeRunAndGetOrThrowWith(rt)
   }

-  def demonitor(ref: Reference): UIO[Unit] = {
+  def demonitor(ref: Reference): Unit = {
     adapter.demonitor(ref.fromScala).unsafeRunWith(adapter.runtime)
   }
 }
@@ -248,7 +248,6 @@ class Process(implicit val adapter: Adapter[_, _]) extends ProcessLike[Adapter[_
     adapter
       .forkScoped(sendZIO(name, msg).schedule(Schedule.spaced(interval)))
       .unsafeRun
-      .getOrThrowFiberFailure()(Unsafe)
   }

   def sendEvery(dest: (RegName, NodeName), msg: Any, delay: Long) = {
@@ -256,7 +255,6 @@ class Process(implicit val adapter: Adapter[_, _]) extends ProcessLike[Adapter[_
     adapter
       .forkScoped(sendZIO(dest, Pid.toScala(self.pid), msg).schedule(Schedule.spaced(interval)))
       .unsafeRun
-      .getOrThrowFiberFailure()(Unsafe)
   }

   def sendAfter(pid: Pid, msg: Any, delay: Long) = {
```

Note that I changed the types for some of the functions because they are supposed to be non-ZIO ones (available for classic Clouseau functions to use) so that their types cannot be wrapped into `ZIO[_, _, _]`.

pgj (approved):

PR #159, opened 2026-04-19
chore(`scalang`): rework `sendEvery` and `sendAfter` in `Service`

pgj:
The definition of `Service#sendEvery` uses `zio.Schedule#spaced` which inserts delays of the given duration between running repeated instances of the effect.  However, this does not consider the time required for running the effect, and it may gradually shift the schedule.  That is why there is [`zio.Schedule#fixed`](https://github.com/zio/zio/blob/b6977b811197397a4c55ca740f175edee9b59e40/core/shared/src/main/scala/zio/Schedule.scala#L1298) which maintains a time-based schedule and inserts delays where the duration has the time required for the effect deducted.  Using that instead will give a more precise implementation.

DRY the definitions of both `sendEvery` and `sendAfter` as part of this change.

### DIFF
--- a/clouseau/src/test/scala/com/cloudant/ziose/clouseau/SendAfterSpec.scala
+++ b/clouseau/src/test/scala/com/cloudant/ziose/clouseau/SendAfterSpec.scala
@@ -1,0 +1,131 @@
+/*
+sbt 'clouseau/testOnly com.cloudant.ziose.clouseau.SendAfterSpec'
+ */
+package com.cloudant.ziose.clouseau
+
+import org.junit.runner.RunWith
+import zio._
+import zio.test.junit.{JUnitRunnableSpec, ZTestJUnitRunner}
+import zio.test.Assertion._
+
+import com.cloudant.ziose.core
+import com.cloudant.ziose.scalang.{Adapter, Pid, Service, ServiceContext, SNode}
+import zio.test._
+import zio.test.TestAspect
+import com.cloudant.ziose.test.helpers.Asserts.containsShapeOption
+import com.cloudant.ziose.test.helpers.TestRunner
+
+class SendAfterService(ctx: ServiceContext[None.type])(implicit adapter: Adapter[_, _]) extends Service(ctx) {
+  var calledArgs: List[Product2[String, Any]] = List()
+
+  override def handleInfo(request: Any): Unit = {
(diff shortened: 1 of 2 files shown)


jiahuili430:
With this PR, if we change `commit_interval_secs` back to 30, will it pass the tests in CI?

pgj:
No, I do not think so.  To me, this seems to be an orthogonal problem.   I only fixed `sendAfter`, but I could not change `sendEvery` in a way that it makes the issue go away with timeouts.  The change for `sendEvery` has a minor impact about making the timing a bit more precise.

scalang/src/main/scala/com/cloudant/ziose/scalang/Service.scala
```diff
@@ -268,41 +267,38 @@ class Process(implicit val adapter: Adapter[_, _]) extends ProcessLike[Adapter[_
   implicit def sym2sendable(to: Symbol): SymSend               = new SymSend(to, this)
   implicit def dest2sendable(dest: (Symbol, Symbol)): DestSend = new DestSend(dest, self, this)
 
+  private def sendEveryCommon(send: UIO[Unit], delay: Long) = {
+    val interval = Duration.fromMillis(delay)
```
jiahuili430: How about using `val interval = delay.millis`?

clouseau/src/test/scala/com/cloudant/ziose/clouseau/SendAfterSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+import zio.test.TestAspect
+import com.cloudant.ziose.test.helpers.Asserts.containsShapeOption
+import com.cloudant.ziose.test.helpers.TestRunner
+
+class SendAfterService(ctx: ServiceContext[None.type])(implicit adapter: Adapter[_, _]) extends Service(ctx) {
+  var calledArgs: List[Product2[String, Any]] = List()
+
+  override def handleInfo(request: Any): Unit = {
+    request match {
+      case 'notify =>
```
jiahuili430: Symbol literals have been deprecated in Scala 2.13, maybe we should use `Symbol("notify")` for longer-term support.

jiahuili430 (approved):
+1

PR #166, opened 2026-04-27
chore(`ci`): work around VM flakiness

pgj:
Requests run in the Elixir and Mango tests tend to time out time to time due to reasons unknown at the moment, but only when they are run in virtualized or containerized environments, such as the CI for this repository, based on GitHub Actions.

Increasing the frequency of recurring commit checks seems to offer a way to mitigate the problem so that at least the resulting flakiness can be eliminated for now.


jiahuili430 (approved):
+1

PR #167, opened 2026-04-30
Make clouseau-ctrl production friendly by providing config file defaults

lacklacklack:
Make clouseau-ctrl production friendly by providing default parameters through a config file.

Related issue: https://github.ibm.com/cloudant/dbcore/issues/659

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,6 +253,17 @@ help:
 		| sort \
 		| awk '{printf("    %-20s", $$1); $$1=$$2=""; print "-" $$0}'
 
+.PHONY: clouseau-ctrl-smoke
+# target: clouseau-ctrl-smoke - Start clouseau1 and verify `clouseau_ctrl service list`
+clouseau-ctrl-smoke: bin/clouseau_ctrl epmd
+	@-cli stop clouseau1 >/dev/null 2>&1
+	@cli start clouseau1 sbt run -Dnode=clouseau1 $(_JAVA_COOKIE)
+	@cli await clouseau1 "$(ERLANG_COOKIE)"
+	@./clouseau-ctrl/_build/default/bin/clouseau_ctrl -config "$(BUILD_DIR)/clouseau-ctrl-config.erl" service list \
+		| tee tmp/clouseau-ctrl-smoke.out \
+		| grep -q analyzer \
+		|| ( echo '>>>>> clouseau_ctrl smoke test failed' ; exit 1 )
+	@-cli stop clouseau1 >/dev/null 2>&1
 
 ############### Tests ###############
 .PHONY: all-tests
--- a/README.md
+++ b/README.md
@@ -119,6 +119,28 @@ JConsole can be connected to a running Clouseau 3.x instance through the standar
 
 ![jmx.png](assets/jmx.png)
 
+## `clouseau_ctrl` configuration
+
+The [`clouseau_ctrl`](./clouseau-ctrl) escript supports default connection settings from a config file in addition to command-line arguments.
+
+Configuration values are looked up in this order (first found wins):
+
+1. Command-line flags: `-name`, `-host`, `-cookie`, `-format`
+2. Environment variables: `CLOUSEAU_NODE_NAME`, `CLOUSEAU_HOST`, `CLOUSEAU_COOKIE`, `CLOUSEAU_FORMAT`
+3. Config file (via `CLOUSEAU_CTRL_CONFIG` or `-config`)
+4. Built-in defaults (e.g., `~/.erlang.cookie` for cookie)
+
+The default config file path is `/etc/clouseau/clouseau-ctrl-config.erl`. A sample file is provided at [`clouseau-ctrl-config.erl`](clouseau-ctrl-config.erl):
+
+```erlang
+{name, "clouseau1"}.
+{host, "127.0.0.1"}.
+%% {cookie, "secret"}.
+{format, table}.
+```
+
+This allows wrappers to provide stable defaults without exposing the Erlang cookie on the process command line.
+
 ## Using `sbt`
 
 The Scala Built Tool `sbt` can be used directly to compile and start up the service, or to run individual unit tests, e.g.
--- a/clouseau-ctrl-config.erl
+++ b/clouseau-ctrl-config.erl
@@ -1,0 +1,6 @@
+{name, "clouseau1"}.
+{host, "127.0.0.1"}.
+%% By default ~/.erlang.cookie is used when cookie is not configured here
+%% or via the CLOUSEAU_COOKIE environment variable.
+%% {cookie, "secret"}.
+{format, table}.
--- a/clouseau-ctrl/src/clouseau_cli.erl
+++ b/clouseau-ctrl/src/clouseau_cli.erl
@@ -5,69 +5,25 @@
     run/2
 ]).
 
+-ifdef(TEST).
+-export([
+    normalize_args/1,
+    read_config/1,
+    resolve_value/3,
+    to_format/1
+]).
+-endif.
+
 parser_options() ->
     #{progname => clouseau_cli}.
 
 spec(Services) ->
+    spec(Services, runtime).
+
+spec(Services, Mode) ->
     #{
         help => "Utility to control 'clouseau' process",
-        arguments => [
-            #{
-                name => script_name,
-                long => "script_name",
-                type => string,
-                default => "clouseau_cli",
-                help => hidden
-            },
-            #{
-                name => script_path,
-                long => "script_path",
-                type => string,
-                default => escript:script_name(),
-                help => hidden
-            },
-            #{
-                name => script_dir,
-                long => "script_dir",
-                type => string,
-                default => filename:dirname(escript:script_name()),
-                help => hidden
-            },
-            #{
-                name => name,
-                short => $n,
-                long => "name",
-                type => string,
-                default => getenv("CLOUSEAU_NODE_NAME"),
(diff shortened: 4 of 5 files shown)


clouseau-ctrl/src/clouseau_cli.erl
```diff
@@ -116,24 +72,205 @@ spec(Services) ->
...
+    EnvValue;
+resolve_value(Value, _EnvValue, _Default) ->
+    Value.
+
+read_config(Path) ->
+    case file:consult(Path) of
+        {ok, [Config]} when is_map(Config) ->
+            Config;
+        _ ->
+            #{}
```
rnewson: _any_ error means we just go with an empty map? I think this could mislead users of the tool. If the file can be missing, then check for that specifically (and if empty map is appropriate in that case, that's the response), but any unexpected error should do something else (e.g, the config file exists but is not readable, we should not proceed)

rnewson:
I had one substantial comment but generally this work is good, though I found some of the functions a bit verbose.

clouseau-ctrl/src/clouseau_cli.erl
lacklacklack: I added a warning log to the _Any case, hopefully this will help avoid confusion on the user side. However I think we should always proceed after this step, leaving it up to the user to decide what configuration to use.

rnewson: I think if the `--config` argument is present and the config file reading fails, that's a fatal error and the command should not be executed. I agree that if it's _not_ present and there's no config file (enoent), that's not an error.

In general, ignoring unexpected errors and then executing a command is a bad idea. It's doing the wrong thing, and a reasonable tool would prevent a user making an avoidable mistake.

rnewson (changes requested):

clouseau-ctrl/src/clouseau_cli.erl
```diff
@@ -116,24 +72,211 @@ spec(Services) ->
         }
     }.
 
+arguments(Mode) ->
```
pgj, line 75: Why do we need the `Mode` argument here?  The `arguments/1` function seems to be called with only a single parameter, which is `runtime`.

clouseau-ctrl/src/clouseau_cli.erl
```diff
@@ -116,24 +72,211 @@ spec(Services) ->
...
+        #{
+            name => script_dir,
+            long => "script_dir",
+            type => string,
+            default => filename:dirname(script_path()),
+            help => hidden
+        },
+        #{
+            name => config,
+            short => $g,
```
pgj: Why `-g` is the short form of this flag?

clouseau-ctrl/src/clouseau_cli.erl
```diff
@@ -116,24 +72,211 @@ spec(Services) ->
...
+
+read_erlang_cookie() ->
+    case os:getenv("HOME") of
         false ->
             undefined;
-        Value ->
-            Value
+        Home ->
+            CookiePath = filename:join(Home, ".erlang.cookie"),
+            case file:read_file(CookiePath) of
```
pgj: I am wondering if that is needed.  If the Erlang cookie is not set explicitly in the program, the run-time system will just find this file automatically?

clouseau-ctrl-config.eterm
```diff
@@ -0,0 +1,8 @@
+#{
+    name => "clouseau1",
+    host => "127.0.0.1",
+    %% By default ~/.erlang.cookie is used when cookie is not configured here
+    %% or via the CLOUSEAU_COOKIE environment variable.
+    %% cookie => "secret",
+    format => table
+}.
```
pgj: There is no newline at the end of the file, please add that.

Makefile
```diff
@@ -253,6 +253,16 @@ help:
 		| sort \
 		| awk '{printf("    %-20s", $$1); $$1=$$2=""; print "-" $$0}'
 
+.PHONY: clouseau-ctrl-smoke
+# target: clouseau-ctrl-smoke - Start clouseau1 and verify `clouseau_ctrl service list`
+clouseau-ctrl-smoke: clouseau-ctrl/_build/default/bin/clouseau_ctrl epmd
+	@set -e; \
+	trap 'scripts/cli stop clouseau1 >/dev/null 2>&1 || true' EXIT; \
```
pgj: Why is this needed?  Other targets do not work like this, but I assume you wanted to implement a similar behavior.

Makefile
```diff
@@ -253,6 +253,16 @@ help:
...
+	scripts/cli start clouseau1 sbt run -Dnode=clouseau1 $(_JAVA_COOKIE); \
+	scripts/cli await clouseau1 "$(ERLANG_COOKIE)"; \
+	output="$$(./clouseau-ctrl/_build/default/bin/clouseau_ctrl -config "$(BUILD_DIR)/clouseau-ctrl-config.eterm" service list)"; \
+	printf '%s\n' "$$output"; \
```
pgj: Maybe it is the [`tee`](https://man7.org/linux/man-pages/man1/tee.1.html) command that you were looking for here?

README.md
```diff
@@ -119,6 +119,30 @@ JConsole can be connected to a running Clouseau 3.x instance through the standar
 
 ![jmx.png](assets/jmx.png)
 
+## `clouseau_ctrl` configuration
+
+The [`clouseau_ctrl`](./clouseau-ctrl) escript supports default connection settings from a config file in addition to command-line arguments.
```
pgj, line 124: "Supports _reading_ default"?

README.md
```diff
@@ -119,6 +119,30 @@ JConsole can be connected to a running Clouseau 3.x instance through the standar
...
+```erlang
+#{
+    name => "clouseau1",
+    host => "127.0.0.1",
+    %% cookie => "secret",
+    format => table
+}.
+```
+
+This allows production wrappers to provide stable defaults without exposing the Erlang cookie on the process command line.
```
pgj: I do not think that we should emphasize "production" here.  This function can come handy in any other environment.

clouseau-ctrl-config.eterm
```diff
@@ -0,0 +1,8 @@
+#{
```
pgj: What is the benefit is using a map here?  Why cannot it just be a series for 2-tuples.  Something similar to that we already have for CouchDB, the format of the `config.erl` file that is generated by the `configure` script there.  In that case, we would not have to bother with indentation and every line could be started in the first column already.

clouseau-ctrl/src/clouseau_cli.erl
lacklacklack: It is also used by the help path. `help/1` builds the spec in help mode so `maybe_default/3` can omit internal runtime placeholder defaults from the generated help.

clouseau-ctrl/src/clouseau_cli.erl
lacklacklack: I agree it is not very intuitive, -c was already taken by -cookie

pgj: In that case perhaps we should not have a short form at all.

clouseau-ctrl/src/clouseau_cli.erl
lacklacklack: You are right, this is redundant. Default cookie handling is already provided by the runtime when we do not set one explicitly. I'm removing this part.

Makefile
lacklacklack: Fair point I will clean this up

clouseau-ctrl-config.eterm
lacklacklack: Changed it to match couchdb style

clouseau-ctrl/src/clouseau_cli.erl
```diff
@@ -116,24 +72,199 @@ spec(Services) ->
...
+        {ok, [{_, _} | _] = Config} ->
+            maps:from_list(
+                [{Key, Value} || {Key, Value} <- Config]
+            );
+        {error, enoent} ->
+            #{};
+        {error, Reason} ->
+            io:format(standard_error, "Error: Unable to read config file ~s: ~p~n", [Path, Reason]),
+            halt(1);
+        {ok, _} ->
```
rnewson, line 262: this won't happen, file:consult will return an error if the file is not erlang terms.

```
1> file:consult("foo.json").
{error,{739,erl_parse,["syntax error before: ",[]]}}
```

lacklacklack: I think that clause is still useful. `file:consult/1` returns an error for input that is not valid Erlang syntax, but it can still return `{ok, Terms}` for syntactically valid input that does not match the expected config shape.

rnewson: ok.

clouseau-ctrl-config.erl
lacklacklack, line 1: change filename extension to `.erl`

Makefile
```diff
@@ -253,6 +253,17 @@ help:
 		| sort \
 		| awk '{printf("    %-20s", $$1); $$1=$$2=""; print "-" $$0}'
 
+.PHONY: clouseau-ctrl-smoke
+# target: clouseau-ctrl-smoke - Start clouseau1 and verify `clouseau_ctrl service list`
+clouseau-ctrl-smoke: clouseau-ctrl/_build/default/bin/clouseau_ctrl epmd
```
lacklacklack: depend on `bin/clouseau_ctrl`

rnewson (approved):

PR #168, opened 2026-05-08
Migrate the toolchain from `asdf+direnv` to `mise`

jiahuili430:
- Use `mise.toml` to specify dependencies and environment variables,
and to load the `bin` and `scripts` directories.
- Leverage `mise` hooks to install dependencies automatically, so
removed the `deps.yaml` file.
- Enable `erlfmt` by specifying the rebar3 plugin in `rebar.config`,
since `mise` can't install `erlfmt` using the GitHub backend.
- Fix `cli zeunit clouseau` command when specifying a cookie.
- Removed code related to `spotbugs` and `yq`, since we do not use them.
- Update scripts and markdown files.


poor-a:
Hey @jiahuili430 
I walked through the README instructions and got this during dev env setup:
```
$ cli verify
checking requirements: 'tput grep cat sed brew awk' ... OK
checking tools: 'type:brew brew:coreutils mise brew' ...
Run the following commands to install missing dependencies
cat: /Users/arturpoor/repo/clouseau/scripts/../.deps: No such file or directory
```
I believe it's not related to your changes in the PR. If it's not a big thing, would it make sense to fix here?

jiahuili430:
Good catch, @poor-a!
Thank you for trying this!

poor-a:
Cool cool, thank you @jiahuili430. It works well now. :)

The PR looks good to me!

README.md
```diff
@@ -17,53 +17,39 @@ enables Clouseau to run on modern JVMs and brings all non-Lucene dependencies up
 Clouseau 3.x is intended to be a drop-in replacement for 2.x in that it remains compatible with existing Clouseau 2.x indexes.
 However, there are some notable changes to the deployment:
 
- * Java 21 is required.
- * `clouseau.ini` is replaced by [app.conf](./app.conf).
- * CouchDB 3.5.0 required.
+ * `Java 21` is required.
```
pgj: Backticks are not needed for either Java 21 or CouchDB 3.5.0.  They are product names not identifiers or file names.

pgj:
For some reason, `mise install` and therefore `eval "$(mise activate zsh)"` does not work for me:

```console
> mise install
mise ERROR scalafmt not found in mise tool registry
mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
```

jiahuili430:
> For some reason, `mise install` and therefore `eval "$(mise activate zsh)"` does not work for me:
> 
> ```
> > mise install
> mise ERROR scalafmt not found in mise tool registry
> mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
> ```

It works for me locally. I tried enabling `zsh` and uninstalling `scalafmt`.
- Restart the terminal, and `mise` will automatically install `scalafmt` when I switch (cd) into the `clouseau` directory.
- Or run `mise install` manually, it will install all the tools.
```bash
jiahuili@Jiahuis-MacBook-Pro ~ % mise --version
...
2026.5.13 macos-arm64 (2026-05-21)

jiahuili@Jiahuis-MacBook-Pro ~ % echo 'eval "$(mise activate zsh)"' >> "${ZDOTDIR-$HOME}/.zshrc"
jiahuili@Jiahuis-MacBook-Pro ~ % cat .zshrc
eval "$(mise activate zsh)"

jiahuili@Jiahuis-MacBook-Pro ~ % mise ls scalafmt
Tool  Version  Source  Requested

# Restart terminal, `mise` will install all the tools automatically.
jiahuili@Jiahuis-MacBook-Pro ~ % cd src/clouseau
scalafmt@3.11.0      verify SLSA provenance

# If not restart the terminal, it needs to manually run `mise install`.
jiahuili@Jiahuis-MacBook-Pro ~ % cd src/clouseau
jiahuili@Jiahuis-MacBook-Pro clouseau % mise install
scalafmt@3.11.0      verify SLSA provenance
```

README.md
jiahuili430: Updated, thanks for the review!

scripts/bootstrap.md
```diff
@@ -1,222 +1,59 @@
...
-
----
-
-## `bash-direnv`: setup `direnv` plugin
-
-**Note**: Keep in mind that the `direnv` plugin can conflict with global installation of `direnv` you might need to remove your existent configuration.
+### Fish
 
-Run following commands to install `direnv` plugin
+If you installed `mise` via `Homebrew`, it activates automatically, so you do not need to perform this step.
```
pgj: Homebrew is not a program name but a product name.  That is, you may want to put `brew` in backticks, because that is the name of the command itself, but not Homebrew, because that is the name of the whole software.

scripts/cli.md
```diff
@@ -9,15 +9,14 @@ Use `cli help <command>` to get more help.
...
 * `bootstrap` - a step-by-step guide to help set up environment
 * `start`     - start clouseau node
 * `stop`      - stop clouseau node
-* `tdump`     - do a java tread dump
+* `tdump`     - capture a Java thread dump
 * `await`     - await the clouseau node
 * `logs`      - get recent logs file for terminated clouseau node
 * `processId` - get clouseau PID
-* `zeunit`    - run `zeunit` tests
+* `zeunit`    - run `ZEUnit` tests
```
pgj: I do not think it is necessary to put ZEunit in backticks.

scripts/cli.md
```diff
@@ -113,16 +106,21 @@ cli processId "clouseau1"
 
 ---
 
-### `zeunit`: Run zeunit tests
+### `zeunit`: Run `ZEUnit` tests
```
pgj: There is no need to put ZEUnit in backticks.

Makefile
```diff
@@ -193,9 +184,9 @@ jmx-prometheus: $(JAR_ARTIFACTS) $(JMX_EXPORTER) epmd
 	@java -javaagent:$(JMX_EXPORTER)=$(JMX_EXPORTER_PORT):$(JMX_EXPORTER_CFG) \
 		-jar $(JAR_ARTIFACTS)
 
-.PHONY: bin/clouseau_ctrl
-bin/clouseau_ctrl: clouseau-ctrl/_build/default/bin/clouseau_ctrl
-	@cp $? $@
+.PHONY: clouseau_ctrl
+clouseau_ctrl: clouseau-ctrl/_build/default/bin/clouseau_ctrl
+	@cp $? $(SCRIPTS_DIR)/$@
```
pgj: `clouseau_ctrl` is not just a helper script that is used for development, but a program that shall be shipped together with the JAR files.  That is, please leave this alone.

mise.toml
```diff
@@ -0,0 +1,27 @@
...
+editorconfig-checker = "3.6.1"
+elixir = "1.19.5-otp-27"
+erlang = "27.3.4.11"
+github-cli = "2.91.0"
+java = "openjdk-21.0.2"
+python = "3.14.4"
+rebar = "3.27.0"
+sbt = "1.12.9"
+scala = "2.13.18"
+scalafmt = "3.11.0"
```
pgj, line 15: There is now version `3.11.1`.

README.md
```diff
@@ -81,35 +68,35 @@ Currently, `cli` provides following commands:
...
 
 If `close_if_idle` is true, then Clouseau will monitor the activity its indexes, and close any with no activity in two consecutive idle check intervals. By default `idle_check_interval_secs` is 300 seconds. which will close an index if it has no activity between 301 to 600 seconds.
 
 ### `-Xmx`
 
-The command line option `-Xmx` sets the maximum heap size for the JVM. The amount of heap usage depends on the number of  open search indexes and also the search load (sorting etc). The amount of heap required usually correlates with the `max_indexes_open` settings, so more indexes open requires more memory.
+The command line option `-Xmx` sets the maximum heap size for the JVM. The amount of heap usage depends on the number of open search indexes and also the search load (sorting etc.). The amount of heap required usually correlates with the `max_indexes_open` settings, so more indexes open requires more memory.
```
pgj: I am not sure if the extra dot after "etc" is needed that much.  We could just abandon that.

README.md
```diff
@@ -81,35 +68,35 @@ Currently, `cli` provides following commands:
...
 
 The recommendation is to set this value to a maximum of one third of the available memory and to never allocate more than 50% of the total available memory. So if the nodes on cluster have 30GB of memory available, then limit `-Xmx` to 10GB and if that's not enough and the user workload still requires more memory then try increasing it to 15GB (50% of available). But be cautious when exceeding 1/3 of the available memory as it could result in less memory available for Erlang runtime and the OS.
 
 ### `-Xms`
 
 This option configures the minimum JVM memory, and is recommended to set in cases of higher maximum heap size (> 8GB). If set, do so at 80% of `-Xmx`. This allows the JVM to set initial memory when Clouseau is started, to avoid dynamic heap resizing and lags.
 
-## How to monitor metrics using `jconsole`
+## Monitor metrics using `jconsole`
```
pgj: If you decided to go with the change, perhaps you may want to say "Monitoring ..." instead.

pgj:
> For some reason, `mise install` [..] does not work for me

Mystery solved: I was using `mise@2026.1.6`.  After upgrading to a more recent version, `2026.5.1.5`, I can confirm that things started to work.

mise.toml
jiahuili430: I have another follow-up PR with `scalafmt`, let's leave it for now.

Makefile
jiahuili430: Will the `bin` folder be shipped with the JAR file? How to verify this?

The `bin` folder exists because we store the `scalafmt` binaries there. After switching to `mise`, we can install and use `scalafmt` directly.
Additionaly, most executable scripts are located in the `scripts` folder—including the restart test. When we `cd` into the `clouseau` directory, `mise` loads this folder, enabling us to use the `cli` commands. That is, if we build `clouseau_ctrl`, it will be automatically activated.

scripts/bootstrap.md
jiahuili430: Updated, thanks for the review!

Makefile
pgj: You can grab the published the `-dist.zip` file for each Clouseau release.  There you can see that `clouseau_ctrl` is included there alongside the (fat) JAR file.

I understand your logic, but the contents of `scripts` is clearly something that one wants to use during the development.  On the contrary, `clouseau_ctrl` makes sense for the users because it is applicable when the application is run (to control or diagnose its behavior) and as such, it should be placed under a different location to avoid the confusion and support the release building process.

Why is not it possible to activate binaries in multiple directories?  On the other hand, please note that users will have to find their own ways to expose `clouseau_ctrl` on the path somehow, they cannot rely on `mise` for that.

jiahuili430: Gotcha, thank you for pointing out the difference. Reverted back.

pgj (approved):

PR #171, opened 2026-05-26
Add `mise` postinstall hook to update tool versions in `mise.toml` and config files

jiahuili430:
When updating tools in `mise.toml`, we need to remember to update the version information within their own respective configuration files simultaneously. For example:

- editorconfig-checker: `.editorconfig-checker.json`
- sbt: `./project/build.properties`
- scalafmt: `.scalafmt.conf`

After adding the `mise` postinstall hook, we can manually update `mise.toml` and then run `mise install`; subsequently, the config file will automatically update to the new version.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -3,16 +3,31 @@
 
 [tools]
 bat = "0.26.0"
-editorconfig-checker = "3.6.1"
+editorconfig-checker = { version = "3.6.1", postinstall = """
+jq --arg v "v$MISE_TOOL_VERSION" '.Version = $v' .editorconfig-checker.json >tmp.json &&
+mv tmp.json .editorconfig-checker.json
+""" }
 elixir = "1.19.5-otp-27"
 erlang = "27.3.4.11"
 github-cli = "2.91.0"
 java = "openjdk-21.0.2"
 python = "3.14.4"
 rebar = "3.27.0"
-sbt = "1.12.9"
+sbt = { version = "1.12.9", postinstall = """
+if [[ $OSTYPE == "darwin"* ]]; then
+  sed -i "" "s/sbt.version=.*/sbt.version=$MISE_TOOL_VERSION/" ./project/build.properties
+else
+  sed -i "s/sbt.version=.*/sbt.version=$MISE_TOOL_VERSION/" ./project/build.properties
+fi
+""" }
 scala = "2.13.18"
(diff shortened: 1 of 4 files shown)


pgj:
Something does not add up with this PR.  While I understand the consistency concerns outlined in the description, I do not see how the PR can provide an adequate answer for that.

The tool-specific configuration files are still under version control, which means that even though `mise` will update them automatically on `install`, one still has to create a commit and open a PR to have those recorded.  That is why we may want to either fully generate those configuration files outside of `git` or inject comments at both places (`mise.toml` and each affected tool configuration file) to remind developers to keep the contents in sync with each other.

jiahuili430:
Thank you for the review, @pgj.

Searched for `sbt` and `mise`, neither offers a built-in feature for generating files based on templates.
While we could potentially achieve this by creating a custom task, but we need to run the task manually.
Adding comments might be the simpler approach to remind us to keep these version numbers synchronized.

Therefore, I added comments in the 3rd commit.
**Note:** `editorconfig-checker` does not support JSONC, so this file was not modified.

pgj:
Does not `mise` have a hook that can verify the version numbers before it is activated?

jiahuili430:
> Does not mise have a hook that can verify the version numbers before it is activated?

Not sure, probably we need to add a custom `enter hook`.

btw, if versions are out of sync, the pre-commit hook will throw an error when we attempt to commit changes to the repository.

mise.toml
```diff
@@ -2,17 +2,36 @@
 # See: https://mise.jdx.dev/configuration.html
 
 [tools]
-bat = "0.26.0"
-editorconfig-checker = "3.6.1"
+bat = "0.26.1"
+# Sync version with ".editorconfig-checker.json"
+editorconfig-checker = { version = "3.7.0", postinstall = """
+jq --arg v "v$MISE_TOOL_VERSION" '.Version = $v' .editorconfig-checker.json >tmp.json &&
```
pgj, line 8: This command assumes the `jq` tool to be available.  If it is not then the post-install script may fail.  I wonder if `mise` could be used to take care of that?

jiahuili430: I've added `jq = "1.8.1"` as a dependency.

pgj: Oh, that is great!

pgj (approved):

PR #173, opened 2026-06-10
chore: update Prometheus JMX Exporter to 1.6.0

pgj:
This [version](https://github.com/prometheus/jmx_exporter/releases/tag/v1.6.0) now includes our extension that makes it possible to customize the endpoint where metrics are exposed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ JMX_EXPORTER_VSN ?= 1.5.0
 JMX_EXPORTER ?= jmx_prometheus_javaagent-$(JMX_EXPORTER_VSN).jar
 JMX_EXPORTER_CFG ?= test/prometheus/jmx_exporter.yaml
 JMX_EXPORTER_PORT ?= 8080
-JMX_EXPORTER_URL := https://github.com/prometheus/jmx_exporter/releases/download/$(JMX_EXPORTER_VSN)/$(JMX_EXPORTER)
+JMX_EXPORTER_URL := https://github.com/prometheus/jmx_exporter/releases/download/v$(JMX_EXPORTER_VSN)/$(JMX_EXPORTER)
 
 comma := ,
 empty :=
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -437,7 +437,7 @@ metrics-tests: $(JAR_ARTIFACTS) $(JMX_EXPORTER) collectd/clouseau.class epmd
 			sort > collectd/metrics.out)
 	@$(call collect_and_compare,\
 			prometheus,\
-			curl -Ss "http://localhost:$(JMX_EXPORTER_PORT)/metrics" | \
+			curl -Ss "http://localhost:$(JMX_EXPORTER_PORT)/custom_path" | \
```
poor-a, line 440: IMO `/metrics` endpoint is more descriptive name than `/custom_path`, which doesn't tell much about the purpose of the endpoint.

pgj: This is just for testing if overriding the default metrics endpoint path works.  This is a new feature in 1.6.0 and it can be misleading if we do not change it to say something else other than the default.

Mind that the Prometheus JMX Exporter integration here is virtual: we do not ship Clouseau that way, the purpose of the integration here is to just to be able to verify if somebody configured this extension in their environment that would still work.

poor-a: Not sure if I follow. If this is for testing, why merge into `main`? It makes more sense to me if the new feature is tested in upstream.

If you insist on merging into `main`, perhaps `/custom_metrics_path` is a better name.

poor-a: > Not sure if I follow. If this is for testing, why merge into main? It makes more sense to me if the new feature is tested in upstream.

Okay, I kind of get it. `prometheus/jmx_exporter.yaml` misled me, I thought that is for production. Maybe the file should be renamed to something indicating that the purpose is testing or providing an example configuration.

poor-a: E.g. `tests/prometheus/jmx_exporter.yaml`.

pgj: Yes, that is a possible option.

pgj: Please check #175 for addressing that part of the review.  I decided to submit that change as a separate PR so we could maintain the original scope of this change.  Note that this PR is now dependent on that other in result.

poor-a (approved):
Looking good

PR #174, opened 2026-06-19
fix(`otp`): make ping configurable for monitoring the CouchDB node

pgj:
It is not recommended to specify very long timeout when calling the [`OtpNode#ping`](https://www.erlang.org/doc/apps/jinterface/assets/java/com/ericsson/otp/erlang/OtpNode.html#ping(java.lang.String,long)) method.  There is only a single attempt is made to connect to the remote node.  This is also a blocking call where it is not so useful to keep it waiting for a long time.

Shorten the ping timeout to 1 second only by default but make it configurable by the user in case fine-tuning is required.

### DIFF
--- a/otp/src/main/scala/com/cloudant/ziose/otp/OTPNodeConfig.scala
+++ b/otp/src/main/scala/com/cloudant/ziose/otp/OTPNodeConfig.scala
@@ -1,18 +1,31 @@
 package com.cloudant.ziose.otp
 
 import com.cloudant.ziose.macros.CheckEnv
-import zio.Config
+import zio.{Config, Duration, durationInt}
 import zio.config.magnolia.deriveConfig
 
-final case class OTPNodeConfig(name: String, domain: String, cookie: Option[String]) {
-  val cookieVal: String = cookie.getOrElse(OTPCookie.findOrGenerateCookie)
+final case class OTPNodeConfig(
+  name: String,
+  domain: String,
+  cookie: Option[String],
+  ping_timeout: Option[Duration],
+  ping_interval: Option[Duration]
+) {
+  val DEFAULT_PING_TIMEOUT: Duration  = 1.seconds
+  val DEFAULT_PING_INTERVAL: Duration = 60.seconds
+
+  val cookieResolved: String         = cookie.getOrElse(OTPCookie.findOrGenerateCookie)
+  val pingTimeoutResolved: Duration  = ping_timeout.getOrElse(DEFAULT_PING_TIMEOUT)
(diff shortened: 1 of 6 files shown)


otp/src/main/scala/com/cloudant/ziose/otp/OTPNodeConfig.scala
```diff
@@ -1,18 +1,31 @@
...
 import zio.config.magnolia.deriveConfig
 
-final case class OTPNodeConfig(name: String, domain: String, cookie: Option[String]) {
-  val cookieVal: String = cookie.getOrElse(OTPCookie.findOrGenerateCookie)
+final case class OTPNodeConfig(
+  name: String,
+  domain: String,
+  cookie: Option[String],
+  ping_timeout: Option[Duration],
+  ping_interval: Option[Duration]
```
poor-a, line 12: Nitpicking: camel case is the common style for scala

```suggestion
  pingTimeout: Option[Duration],
  pingInterval: Option[Duration]
```

otp/src/main/scala/com/cloudant/ziose/otp/OTPNodeConfig.scala
```diff
@@ -1,18 +1,31 @@
...
+) {
+  val DEFAULT_PING_TIMEOUT: Duration  = 1.seconds
+  val DEFAULT_PING_INTERVAL: Duration = 60.seconds
+
+  val cookieVal: String         = cookie.getOrElse(OTPCookie.findOrGenerateCookie)
+  val pingTimeoutVal: Duration  = ping_timeout.getOrElse(DEFAULT_PING_TIMEOUT)
+  val pingIntervalVal: Duration = ping_interval.getOrElse(DEFAULT_PING_INTERVAL)
```
poor-a: Nitpicking: feels like there's better naming for the *Val variables. Maybe


```suggestion
  val cookieResolved: String         = cookie.getOrElse(OTPCookie.findOrGenerateCookie)
  val pingTimeoutResolved: Duration  = ping_timeout.getOrElse(DEFAULT_PING_TIMEOUT)
  val pingIntervalResolved: Duration = ping_interval.getOrElse(DEFAULT_PING_INTERVAL)
```

poor-a (approved):
LGTM!

otp/src/main/scala/com/cloudant/ziose/otp/OTPNodeConfig.scala
poor-a: Please ignore.

PR #175, opened 2026-06-20
chore(`test`): move all the test-related files under the `test` directory

pgj:
Over time we accumulated a number of configuration and auxiliary files to stage certain test scenarios in the repository root. This may give the impression that these are actually included in the release due to their location.

Make the distinction and role clear by moving such files under the already existing `test` sub-directory, where they belong to.

Thanks @poor-a for calling this out.


poor-a (approved):
LGTM!

PR #178, opened 2026-06-29
build: enable native access for Lucene core functions

pgj:
[Leveraging memory-mapped I/O for Lucene 10 requires the invocation of the `madvise()` C library function](https://lucene.apache.org/core/10_4_0/core/org/apache/lucene/store/MMapDirectory.html), which is not unconditionally accessible but only when that is explicitly allowed for the Java Virtual Machine.

Enable that for developer and continuous integration workflows since that is something that one may want to do in production anyway.  While here, merge the injection of the required flag with how Erlang cookies are passed down to the application.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,16 +24,18 @@ TIMEOUT_ELIXIR_SEARCH?=20m
 
 REBAR?=rebar3
 
+JVM_FLAGS= \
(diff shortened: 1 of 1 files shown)


rnewson (approved):

poor-a (approved):
Looking good!

PR #180, opened 2026-07-06
Improve logging and code quality

poor-a:
Motivation: logs are essential to monitor services and tracking down bugs. Changes in this PR are made while investigating a race condition, and they contributed to finding the root cause. See https://github.com/apache/couchdb/pull/6059

This PR adds logging to some hotspots of the code: message handling, index service, main.

### DIFF
--- a/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
+++ b/clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
@@ -80,6 +80,7 @@ case object InvalidReader extends Exception
 class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adapter[_, _]) extends Service(ctx) with Instrumented {
   import IndexService._
 
+  logger.debug(s"Started index service with path ${ctx.args.writer.getDirectory}")
   var lazyReader: Option[DirectoryReader] = None
   var updateSeq = getCommittedSeq
   var pendingSeq = updateSeq
@@ -337,6 +338,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
 
   private def search(request: SearchRequest): Any = {
     val queryString = request.options.getOrElse('query, "*:*").asInstanceOf[String]
+    logger.debug(s"Evaluating query: ${queryString}")
     val refresh = request.options.getOrElse('refresh, true).asInstanceOf[Boolean]
     val limit = request.options.getOrElse('limit, 25).asInstanceOf[Int]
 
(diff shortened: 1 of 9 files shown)


clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -337,6 +338,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
 
   private def search(request: SearchRequest): Any = {
     val queryString = request.options.getOrElse('query, "*:*").asInstanceOf[String]
+    logger.debug(s"Evaluating query: ${queryString}")
```
pgj, line 341: Should not this be something that is logged on the "trace" level?  Looks like we do not yet have a `LoggerFactory#trace` method (but a `TRACE` log level)  though.  This might be the right time to introduce that too.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/IndexService.scala
```diff
@@ -352,6 +354,7 @@ class IndexService(ctx: ServiceContext[IndexServiceArgs])(implicit adapter: Adap
 
     parseQuery(queryString, partition) match {
       case baseQuery: Query =>
+        logger.debug(s"Parsed query: $baseQuery")
```
pgj, line 357: I think this is more meant to be for the `trace` level.

core/src/main/scala/com/cloudant/ziose/core/Actor.scala
```diff
@@ -101,8 +101,12 @@ class AddressableActor[A <: Actor, C <: ProcessContext](actor: A, context: C)
   }
 
   def onMessage(message: MessageEnvelope)(implicit trace: Trace): UIO[ActorResult] = {
-    def onFunc = actor.onMessage(message, ctx)
-    callHandler(onFunc, ActorCallback.OnMessage)
+    lazy val onFunc = actor.onMessage(message, ctx)
+    for {
+      _   <- ZIO.logDebug(s"Handling mail message: $message")
```
pgj, line 106: That is a pity that ZIO does not have a `logTrace` method.  Logging all instances of `onMessage` would probably generate a *lot* of messages.  The reason why I am wary about using the `DEBUG` log level for emitting such information because by default, the CI uses that level on running the tests.

macros/src/main/scala/com/cloudant/ziose/macros/Version.scala
```diff
@@ -0,0 +1,19 @@
+package com.cloudant.ziose.macros
+
+import scala.language.experimental.macros
+import scala.reflect.macros.blackbox
+
+object Version {
```
pgj, line 6: I do not understand why this version of getting version information is any better that we already have in the code base.  Please note that Clouseau can return its version already via EDP there is no need to reinvent the wheel here...?

PR #172, opened 2026-06-01
Package release in tar.gz

poor-a:
The PR adds `.tar.gz` format to Clouseau release artefacts alongside the current `.zip` format.

PR #169, opened 2026-05-13
Upgrade JInterface to OTP-27

jiahuili430:
- Upgrade Erlang to `27.3.4.11` and Elixir to `1.19.5-otp-27`.
- Update JInterface to `1.14.1 (OTP-27)`: Identify the differences between JInterface `1.13.2 (OTP-25)` and `1.14.1 (OTP-27)`, and update the codebase in vendor.

PR #162, opened 2026-04-22
Enable manually trigger workflows

jiahuili430:

PR #181, opened 2026-07-08
release: 3.4.0-RC1

pgj:
Changes since 3.3.0:

- #155
- #156
- #157
- #158
- #159
- #162
- #166
- #167
- #168
- #169
- #170
- #171
- #172
- #173
- #174
- #175

Note that this is a Release Candidate because we will still need to do some preliminary testing on production systems before we could consider the contents final.


poor-a (approved):
Seems good!

PR #183, opened 2026-07-29
Parse Clouseau config parameters from system properties

poor-a:
A regression was identified in Clouseau 3 compared to Clouseau 2. Version 3 doesn't consider Java system properties at startup when it resolves its configuration. The exception is `-Dnode`, which allows one to index into the `config` list of the configuration file. 

That is especially problematic if client code sets the working directory through system properties. Multiple Clouseau instances will race to lock the same working directory.

Apart from fixing the regression, a secondary task was to revamp the configuration handling code.

The list of changes:
  - Configuration parser library is changed from `magnolia` to combination of [`typesafe`](https://github.com/lightbend/config) and [`pureconfig`](https://github.com/pureconfig/pureconfig)
  - Default parameter values scattered across the code base are moved into their respective case classes in `Configuration.scala`. This places them in the same file for easier overview and allows for easier class instantiation in production code and tests
  - Added configuration logging at startup to verify that Clouseau parsed its configuration correctly
  - New tests are added for application bootstrapping with and without configuration file, consuming system properties
  - Existing configuration tests are broken down into multiple test configuration files instead of programatically generating configuration strings. That allows us to test the file parser being actually used by the production code


clouseau/src/main/scala/com/cloudant/ziose/clouseau/Main.scala
```diff
@@ -47,9 +24,9 @@ object Main extends ZIOAppDefault {
         workerCfg.node.pingIntervalResolved
       )
       worker <- ZIO.service[EngineWorker]
-      logLevel = loggerCfg.level.getOrElse(LogLevel.Debug)
+      logLevel = loggerCfg.level.getOrElse(LogLevel.Info)
```
pgj: Why is there a change in the default?  There is no justification for adjusting that.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Main.scala
```diff
@@ -58,42 +35,48 @@ object Main extends ZIOAppDefault {
...
+        .scoped(main(nodeConfig, metricsRegistry, loggerCfg))
+        .provide(OTPLayers.nodeLayers(engineId, workerId, otpConfig))
     } yield ()
   }
 
+  private def patchConfig(config: Configuration): UIO[Configuration] =
+    AppCfg.patchClouseauConfig(config.clouseau) match {
+      case Left(err) =>
+        for {
+          _ <- ZIO.logDebug(s"Parse error while reading system properties: $err - ignoring system properties")
```
pgj: Is this the right choice not stopping immediately on encountering parse errors for the system properties?  I can understand that it is worth logging instead of silently ignoring them.  But this is just a debug message which will not be visible on higher log levels at all (which is common in production, for example).  

I am wary that (basically) silently dropping ill-formed values will just confuse the users.  I would rather make the application stop in such cases, like we do for the ill-formed configuration values.  There shall be no differences in treating those in my opinion.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Main.scala
```diff
@@ -58,42 +35,48 @@ object Main extends ZIOAppDefault {
...
-      loggerCfg       = appCfg.logger
+      appCfg    <- ZIO.service[AppCfg]
+      workerCfg <- patchConfig(appCfg.config(appCfg.configIndex))
+      loggerCfg = appCfg.logger
+      _ <- ZIO.logInfo(s"Resolved configuration: ${appCfg.copy(config = List(workerCfg))}")
       metricsRegistry = ClouseauMetrics.makeRegistry
       metricsLayer    = ClouseauMetrics.makeLayer(metricsRegistry)
       _ <- ZIO
-        .scoped(app("Main", workerCfg, metricsRegistry, loggerCfg))
+        .scoped(app("Main", workerCfg, metricsRegistry, loggerCfg)) // TODO
```
pgj: TODO what?

clouseau/src/test/scala/com/cloudant/ziose/clouseau/ConfigSpec.scala
```diff
@@ -6,163 +6,124 @@ package com.cloudant.ziose.clouseau
...
-          ) ?? "Expect error to be for 'config.[0].capacity.${key}' path"
-          && assertTrue(
-            error.exists(
-              _.toString.contains(s"Exponent must be greater than 0 and less than or equal to 16 (got '$capacity'")
+  def spec: Spec[Any, Nothing] = {
+    suite("ConfigSpec")(
+      suite(s"configSuite for exponent type")(
+        test(s"Ensure we can parse lower bound")(
+          for {
+            config <- AppCfg.fromHoconFile("src/test/resources/testCapacity_LowerBound.conf")
```
pgj, line 21: Why cannot we just stick to the original solution that creates the configuration files in memory using a template on demand, which does not require separate files for covering each of the cases?

core/src/main/scala/com/cloudant/ziose/core/Exponent.scala
```diff
@@ -1,6 +1,6 @@
 package com.cloudant.ziose.core
 
-final case class Exponent(private val value: Int) extends AnyVal {
+final case class Exponent(val value: Int) extends AnyVal {
```
pgj: Why is the `private` modifier removed?  This parameter seems to be an internal detail to me.

build.sbt
```diff
@@ -165,7 +165,9 @@ lazy val commonSettings = Seq(
     "dev.zio"       %% "zio-test"                          % versions("zio") % Test,
     "dev.zio"       %% "zio-test-junit"                    % versions("zio") % Test,
     "com.github.sbt" % "junit-interface"                   % "0.13.3"        % Test,
-    "junit"          % "junit"                             % "4.13.2"        % Test
+    "junit"          % "junit"                             % "4.13.2"        % Test,
+    "com.typesafe"   % "config"                            % "1.4.9",
+    "com.github.pureconfig" %% "pureconfig"                % "0.17.10"
```
pgj: Perhaps it would make sense to align the `%%` and `%` symbols with the rest of the dependencies.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Configuration.scala
```diff
@@ -42,106 +71,66 @@ object LogConfiguration {
       case "NONE"    => Right(LogLevel.None)
       case _         =>
         Left(
-          Error.InvalidData(message = {
-            s"LogLevel must be one of (case insensitive) ALL|FATAL|ERROR|WARNING|INFO|DEBUG|TRACE|NONE (got '${value}')"
-          })
+          pureconfig.error.CannotConvert(
+            value,
+            "LogLevel",
+            "LogLevel must be one of (case insensitive) ALL|FATAL|ERROR|WARNING|INFO|DEBUG|TRACE|NONE"
```
pgj, line 77: I think it is still mentioning the originally parsed value (the part saying "(got '${value}')") for debugging purposes.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Configuration.scala
```diff
@@ -176,51 +167,130 @@ final case class CapacityConfiguration(
...
+  private val fieldCase: ConfigFieldMapping = ConfigFieldMapping(SnakeCase, SnakeCase)
+  implicit val capacityConfigurationSyntaxOptions: ProductHint[CapacityConfiguration] = ProductHint(fieldCase)
+  implicit val propertyConfigurationSyntaxOptions: ProductHint[PropertyConfiguration] = ProductHint(fieldCase)
+  implicit val clouseauConfigurationSyntaxOptions: ProductHint[ClouseauConfiguration] = ProductHint(fieldCase)
+  implicit val appCfgSyntaxOptions: ProductHint[AppCfg]                               = ProductHint(fieldCase)
 
   private val DEFAULT_CFG: String = "app.conf"
 
-  def layer: ZLayer[ZIOAppArgs, Config.Error, AppCfg] = {
+  val conf: com.typesafe.config.Config = ConfigFactory.parseFile(new java.io.File(DEFAULT_CFG))
```
pgj: Is this line going to always assume that the `app.conf` is present in the directory where the application is started?  What does guarantee that the name of the configuration file could be passed as a command-line argument?

clouseau/src/test/scala/com/cloudant/ziose/clouseau/ConfigSpec.scala
poor-a: I believe it is easier to follow the test code: now it's just a series of `test` statements, while the test input is also more straightforward. The main reason is that this way we can test the production code actually in use. The previous `fromHoconString` method was only for testing purposes, while the `fromHoconFile` used by `main` wasn't tested here.

core/src/main/scala/com/cloudant/ziose/core/Exponent.scala
poor-a: Good point. I thought public access is necessary to convert `Exponent` back to JSON format, and that's not the case.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Configuration.scala
poor-a: The new version also prints the original not parsed value in the 1st parameter. Is it sufficient or there is a better version? Not sure what the ask is here.

build.sbt
poor-a: Done.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Main.scala
poor-a: Removed!

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Main.scala
poor-a: `Info` level is the default in most cases I believe. I don't have strong preference.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Configuration.scala
poor-a: Done. If file is not specified as argument then Clouseau starts with default parameters hard coded.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Main.scala
pgj: Let us keep it like it was originally.  This is completely out of the scope of this PR.

clouseau/src/main/scala/com/cloudant/ziose/clouseau/Configuration.scala
pgj: Okay, that is fine then.  The "ask" here is to maintain the same semantics and visibility into the details, such as the value of the non-parsed input.

pgj:
Great work!  But please note I would postpone merging this PR until 3.4.0 is out.  I would not want to add more changes to that release, it is already packed with a few.

PR #185, opened 2026-08-04
fix(`ci`): unbreak + resiliency enhancements

pgj:
The CI stopped working (at the `metrics-tests` phase) because the upstream for the Prometheus JMX Exporter adjusted their tag for the 1.6.0 release retroactively.  Unfortunately, the errors coming from this change were not apparent.  So, on top of addressing the root cause, the following enhancements have been implemented to make it easier to identify similar problems in the future.

- There was an oversight in 74e2f663 where the directory names have not been adjusted for the calling the `collect_and_compare` function.  The issue was not spotted before because the command did not raise any error when `diff` could not find the files to compare.

- The (optional) Java cookie was not passed for the application in `metrics-tests`.

- Make `curl` fail for HTTP 404/500 responses explicitly, because, by default, this is not done.  This let `curl` to create a file with the `Not Found` response for the Java Agent JAR, which broke the application start-up process.

- Make the uses of `cli await` present the application log on failures on startup so that one could immediately see if the problem was related to that.

- Revamp the smoke test for `clouseau-ctrl` and make it part of the CI pipeline, because it has not been hooked up apparently.

- Externalize the assumed locale settings to make the behavior of `sort` and `grep` deterministic across different platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,7 +177,7 @@ jar: $(ARTIFACTS_DIR)/$(JAR_PROD)
 jartest: $(ARTIFACTS_DIR)/$(JAR_TEST)
 
 $(JMX_EXPORTER):
-	@curl -LSso $(JMX_EXPORTER) $(JMX_EXPORTER_URL)
+	@curl -fLSso $(JMX_EXPORTER) $(JMX_EXPORTER_URL)
 
 .PHONY: jmx-prometheus
 # target: jmx-prometheus - Export metrics to Prometheus
--- a/Makefile
+++ b/Makefile
@@ -432,7 +432,8 @@ metrics-tests: $(JAR_ARTIFACTS) $(JMX_EXPORTER) test/collectd/clouseau.class epm
 			-Dcom.sun.management.jmxremote.ssl=false \
 			-Dcom.sun.management.jmxremote.password.file=jmxremote.password \
 			-javaagent:$(JMX_EXPORTER)=$(JMX_EXPORTER_PORT):$(JMX_EXPORTER_CFG) \
-			-jar $(JAR_ARTIFACTS) test/conf/metrics.app.conf > /dev/null
+			$(_JAVA_COOKIE) \
+			-jar $(JAR_ARTIFACTS) test/conf/metrics.app.conf
 	@sleep 5
 	@cli await $(node_name) "$(ERLANG_COOKIE)"
 	@echo "Warming up Clouseau to expose all the metrics"
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 # Use bash to enable `read -d ''` option
 SHELL := /bin/bash
 
+export LC_ALL := C
+
 BUILD_DIR=$(shell pwd)
 ARTIFACTS_DIR=$(BUILD_DIR)/artifacts
 CI_ARTIFACTS_DIR=$(BUILD_DIR)/ci-artifacts
--- a/Makefile
+++ b/Makefile
@@ -251,7 +251,7 @@ help:
 clouseau-ctrl-smoke: bin/clouseau_ctrl epmd
 	@-cli stop clouseau1 >/dev/null 2>&1
 	@cli start clouseau1 sbt run -Dnode=clouseau1 $(_JAVA_COOKIE)
-	@cli await clouseau1 "$(ERLANG_COOKIE)"
+	@cli await clouseau1 "$(ERLANG_COOKIE)" || $(MAKE) await-failed ID=$@
 	@./clouseau-ctrl/_build/default/bin/clouseau_ctrl -config "$(BUILD_DIR)/clouseau-ctrl-config.erl" service list \
 		| tee tmp/clouseau-ctrl-smoke.out \
 		| grep -q analyzer \
@@ -277,7 +277,7 @@ FORCE: # https://www.gnu.org/software/make/manual/html_node/Force-Targets.html
 zeunit: $(ARTIFACTS_DIR)/$(JAR_TEST) epmd FORCE
 	@cli start $@ "java $(_JAVA_COOKIE) -jar $<"
 	@sleep 5
-	@cli await $(node_name) "$(ERLANG_COOKIE)"
+	@cli await $(node_name) "$(ERLANG_COOKIE)" || $(MAKE) await-failed ID=$@
 	@cli zeunit $(node_name) "$(EUNIT_OPTS)" || $(MAKE) test-failed ID=$@
 	@$(call to_artifacts,zeunit,test-reports)
 	@cli stop $@
@@ -285,7 +285,7 @@ zeunit: $(ARTIFACTS_DIR)/$(JAR_TEST) epmd FORCE
 concurrent-zeunit-tests: $(ARTIFACTS_DIR)/$(JAR_TEST) epmd FORCE
 	@cli start $@ "java $(_JAVA_COOKIE) -jar $< test/conf/concurrent.app.conf"
 	@sleep 5
-	@cli await $(node_name) "$(ERLANG_COOKIE)"
+	@cli await $(node_name) "$(ERLANG_COOKIE)" || $(MAKE) await-failed ID=$@
 	@cli zeunit $(node_name) "$(EUNIT_OPTS)" || $(MAKE) test-failed ID=$@
 	@cli stop $@
 
@@ -304,7 +304,7 @@ syslog-test: $(JAR_ARTIFACTS) epmd FORCE
 		-e "s/%%LEVEL%%/$(LEVEL)/" \
 		test/conf/syslog.app.conf.templ > test/conf/syslog.app.conf
 	@cli start $@ "java -jar $< test/conf/syslog.app.conf"
-	@cli await $(node_name) "$(ERLANG_COOKIE)" || $(MAKE) test-failed ID=$@
+	@cli await $(node_name) "$(ERLANG_COOKIE)" || $(MAKE) await-failed ID=$@
 	@echo ">>> Waiting for Clouseau to generate logs (5 seconds)"
 	@sleep 5
 	@cli stop $@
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -402,7 +408,7 @@ define collect_and_compare
 	echo -e "\nCollecting $(1) metrics"
 	$(2)
 	echo "Comparing $(1) metrics with expectations:"
-	DIFF=$$(diff -u $(1)/metrics.out $(1)/metrics.expected); \
```
poor-a, line 405: I'm not sure about this one. Assignment of `DIFF` and reference to `DIFF` run in separate subshells, with the second subshell not having access to actual value of `DIFF`

Makefile
```diff
@@ -402,7 +408,7 @@ define collect_and_compare
...
+	DIFF=$$(diff -u $(1)/metrics.out $(1)/metrics.expected)
 	if [[ -z $$DIFF ]]; then \
 		echo -e "\tEverything is in order"; \
 	else \
```
poor-a: I believe it should be

```suggestion
	DIFF=$$(diff -u $(1)/metrics.out $(1)/metrics.expected); \
	if [[ $$? -eq 0 ]]; then \
		echo -e "\tEverything is in order"; \
	else \
```

Makefile
```diff
@@ -277,15 +277,15 @@ FORCE: # https://www.gnu.org/software/make/manual/html_node/Force-Targets.html
 zeunit: $(ARTIFACTS_DIR)/$(JAR_TEST) epmd FORCE
 	@cli start $@ "java $(_JAVA_COOKIE) -jar $<"
 	@sleep 5
-	@cli await $(node_name) "$(ERLANG_COOKIE)"
+	@cli await $(node_name) "$(ERLANG_COOKIE)" || $(MAKE) await-failed ID=$@
```
poor-a, line 282: Just for the record, it's confusing that `await` awaits on `$(node_name) == clouseau1` and `await-failed` takes  the process name `$@ == zeunit` as input.

Makefile
```diff
@@ -251,7 +251,7 @@ help:
 clouseau-ctrl-smoke: bin/clouseau_ctrl epmd
 	@-cli stop clouseau1 >/dev/null 2>&1
 	@cli start clouseau1 sbt run -Dnode=clouseau1 $(_JAVA_COOKIE)
-	@cli await clouseau1 "$(ERLANG_COOKIE)"
+	@cli await clouseau1 "$(ERLANG_COOKIE)" || $(MAKE) await-failed ID=$@
```
poor-a: `cli start clouseau1` sets process name `ID=clouseau1` but the await-failed assumes `ID=clouseau-ctrl-smoke` if I'm not mistaken.

```
@cli start $@ sbt run -Dnode=clouseau1 $(_JAVA_COOKIE)
```

poor-a:
The file `test/collectd/metrics.expected` is not sorted at the moment. So the `metrics-tests` always fails. Similarly `test/prometheus/metrics.expected` needs sorting.

poor-a (changes requested):

Makefile
pgj: Hrm, I was just following the existing pattern.  Maybe the original code is also faulty?

pgj: If I am not reading it wrong, the `node_name` is always going to be `clouseau1`.  Yet `cli show logs` could still find the logs associated with the node that was shut down.

pgj:
`metrics-tests` does not fail for me because the generated `.out` file contains the metrics in the same order.  So it is not always happening, but I can understand that sometimes it may happen, since the ordering is not guaranteed by the code (before comparing).

pgj:
Actually it is sorted.  Please check the call sites for the `collect_and_compare` function and you will find the `sort` command there.

Makefile
pgj: We can always group them into a single sub-shell if that is a concern.  Let me validate that.

Makefile
pgj: Ah, now I think I understand: `cli show logs` will get you the latest instance of the same id (with the latest timestamp) and nothing if there is no log with that id.  Quite silent failure, so everything is possible there.  Nice find!

pgj: Another (related) discovery: `cli start` and `cli stop` uses the name to create a separate file for the test, a prefix for the resulting logs.  That is what `cli show logs` also uses.  On the contrary, `cli await` uses the name to access the (virtual) Erlang node (of Clouseau) which is basically `node_name`, i.e. `clouseau1` all the time.  That is confusing...

pgj: With that said (see my comment above), I think we are good here?

Makefile
pgj: That is more complicated than that, see my comment in the other thread.  In summary, I think it is okay as much as it could be at the moment.

pgj: Nevertheless I decided to tweak the ids a bit, just to make this test case consistent with others and reduce the factor of confusion.

Makefile
pgj: Yes, you are right.  I have phrased this differently my fix, but the problem is this.

poor-a:
Okay, interesting. This is what I got:

```
OK (skipped=1)

Collecting  test/collectd metrics
Comparing  test/collectd metrics with expectations:
	FAILED: Metrics is different from " test/collectd/metrics.expected"!
--- test/collectd/metrics.out	2026-08-05 14:06:07
+++ test/collectd/metrics.expected	2026-07-29 15:32:48
@@ -1,3 +1,4 @@
+"com.cloudant.clouseau":name="NativeFSLock.count",type="IndexManagerService"
 "com.cloudant.clouseau":name="commits",type="IndexService"
 "com.cloudant.clouseau":name="deletes",type="IndexService"
 "com.cloudant.clouseau":name="indexes.seen.1d",type="IndexManagerService"
@@ -6,7 +7,6 @@
 "com.cloudant.clouseau":name="indexes.seen.1w",type="IndexManagerService"
 "com.cloudant.clouseau":name="lru.evictions",type="IndexManagerService"
 "com.cloudant.clouseau":name="lru.misses",type="IndexManagerService"
-"com.cloudant.clouseau":name="NativeFSLock.count",type="IndexManagerService"
 "com.cloudant.clouseau":name="opens",type="IndexManagerService"
 "com.cloudant.clouseau":name="partition_search.timeout.count",type="IndexService"
 "com.cloudant.clouseau":name="searches",type="IndexService"
make: *** [metrics-tests] Error 1
```

poor-a:
After sorting `metrics.expected` that error is gone.

pgj:
There is something I can think about.  Maybe you are using an implementation of `sort` that is different from what the CI (and myself) use?

Mind that on macOS, the default implementation is coming from FreeBSD and that is the so-called "BSD sort".  While, on the other hand, there is "GNU sort" which common to GNU/Linux systems (such as Ubuntu used for the runner in GitHub Actions), and it may be possible that they produce different results in certain cases.

For example, consider the following:

```
% sort --version
2.3-Apple (195)
% gsort --version
sort (GNU coreutils) 9.6
Copyright (C) 2025 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and Paul Eggert.
```

Makefile
poor-a: > On the contrary, cli await uses the name to access the (virtual) Erlang node (of Clouseau) which is basically node_name, i.e. clouseau1 all the time. That is confusing...

That's exactly what I meant. Thanks!
Yes, that concern is not related to this PR. Just wanted to highlight that is a nice-to-have task for someone having spare cycles.

poor-a:
I have slightly different versions:
```
$ sort --version
2.3-Apple (199)
$ gsort --version
sort (GNU coreutils) 9.11
Copyright (C) 2026 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and Paul Eggert.
```

I'm on macos 26.6

poor-a:
Ah, I found the reason: the locale setting
```
$ LC_ALL=C.UTF-8 sort test/collectd/metrics.expected 
"com.cloudant.clouseau":name="NativeFSLock.count",type="IndexManagerService"
"com.cloudant.clouseau":name="commits",type="IndexService"
"com.cloudant.clouseau":name="deletes",type="IndexService"
"com.cloudant.clouseau":name="indexes.seen.1d",type="IndexManagerService"
"com.cloudant.clouseau":name="indexes.seen.1h",type="IndexManagerService"
"com.cloudant.clouseau":name="indexes.seen.1m",type="IndexManagerService"
"com.cloudant.clouseau":name="indexes.seen.1w",type="IndexManagerService"
"com.cloudant.clouseau":name="lru.evictions",type="IndexManagerService"
"com.cloudant.clouseau":name="lru.misses",type="IndexManagerService"
"com.cloudant.clouseau":name="opens",type="IndexManagerService"
"com.cloudant.clouseau":name="partition_search.timeout.count",type="IndexService"
"com.cloudant.clouseau":name="searches",type="IndexService"
"com.cloudant.clouseau":name="updates",type="IndexService"
```
versus
```
$ LC_ALL=en_US.UTF-8 sort test/collectd/metrics.expected 
"com.cloudant.clouseau":name="commits",type="IndexService"
"com.cloudant.clouseau":name="deletes",type="IndexService"
"com.cloudant.clouseau":name="indexes.seen.1d",type="IndexManagerService"
"com.cloudant.clouseau":name="indexes.seen.1h",type="IndexManagerService"
"com.cloudant.clouseau":name="indexes.seen.1m",type="IndexManagerService"
"com.cloudant.clouseau":name="indexes.seen.1w",type="IndexManagerService"
"com.cloudant.clouseau":name="lru.evictions",type="IndexManagerService"
"com.cloudant.clouseau":name="lru.misses",type="IndexManagerService"
"com.cloudant.clouseau":name="NativeFSLock.count",type="IndexManagerService"
"com.cloudant.clouseau":name="opens",type="IndexManagerService"
"com.cloudant.clouseau":name="partition_search.timeout.count",type="IndexService"
"com.cloudant.clouseau":name="searches",type="IndexService"
"com.cloudant.clouseau":name="updates",type="IndexService"
```
My shell is configured to `LC_ALL=en_US.UTF-8` while the CI probably has `LC_ALL=C.UTF-8 `

pgj:
Ah, cool.  Let us just make it explicit to avoid any further mishaps.

poor-a (approved):

PR #192, opened 2026-08-16
fix(`ci`): cover documentation

pgj:
Documentation was sitting in the `docs` sub-directory but we did not have a coverage for that in the CI.  Create a separate workflow that could be run independently and therefore in parallel to regular build workflow (when required) to address this gap, based on the existing `docs` `make` target.

Make the documentation officially part of the releases.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,0 +1,31 @@
+name: Build documentation
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'docs/**'
+      - '.github/workflows/docs.yml'
+      - 'mise.toml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  build:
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v6
+      - run: git config advice.detachedHead false
+      - name: Install mise & tools
+        uses: jdx/mise-action@v4
+
+      - name: Build documentation
+        run: mise exec -- make docs
```
poor-a: The `mise exec --` part can be omitted once https://github.com/cloudant-labs/clouseau/pull/195 is merged

poor-a (approved):
Looks good. Care should be taken to unify with https://github.com/cloudant-labs/clouseau/pull/195

PR #193, opened 2026-08-16
fix(`zeunit`): tighten the requirements on version reporting over EDP

pgj:
Instead of relying on a regular expression to verify if the `clouseau_rpc:version/0` call returns some semantic version number, use the Scala project version, as given in `version.sbt`, to check if that is the one that is returned actually.

This can help to avoid unsuspiciously going by the assumption that the way how the version number is extracted in Scala / Java via [`getClass.getPackage.getImplementationVersion`][1] would always report the right value.  Because, according to the documentation, there is no actual guarantee about that:

> It consists of any string assigned by the vendor of this implementation and does not have any particular syntax specified or expected by the Java runtime.

Thanks @poor-a for pointing this out.

[1]: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Package.html#getImplementationVersion()

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ endif
 ifeq ($(PROJECT_VSN),)
 	# technically we could use 'sbt -Dsbt.supershell=false -error "print version"'
 	# but it takes 30 seconds to run it. So we go with direct access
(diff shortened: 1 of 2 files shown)


poor-a (approved):
Looks good, thanks!

PR #195, opened 2026-08-17
fix(`ci`): unbreak & unify

pgj:
This pull request features two changes to the CI integration:

- _Set Java home explicitly for SBT._  There may be configurations where `mise` fails to properly override the Java version in use for `sbt`, which can lead to problems with the build.  A notable example of that is the `ubuntu-latest` GitHub Actions runner where Java 25 is available, while `mise` wants to use Java 21 and `sbt` somehow still falls back to the former.  This way the build may be unbroken.

- _Revamp and unify workflows._ The `jdx/mise-action` GitHub Action does not require to call `mise exec` on every step, but with the `env` and `export_path` settings, it can just propagate the necessary environment variables.  In addition to that, rename the workflows to omit the references to the major version number, as they could be re-used for testing the code and producing artifacts for the 4.x line of releases too.  Also, remove the single testing step from the release workflow as we could assume that extensive testing has happened before, on merging changes for the `main` branch.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,4 +1,4 @@
-name: Build 3.x
+name: Build & Test
 
 on:
   pull_request:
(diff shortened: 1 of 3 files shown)


poor-a (approved):
LGTM

PR #197, opened 2026-08-22
chore(`docs`): showcase how to run the application

pgj:
The documentation does not discuss how the artifact should be run outside using SBT, even though it touches questions about how the configuration parameters should be tuned.  To address this gap, add a section with a suggestion on how the application should be started up.

In line with that, some other changes were made:

- _Rename `app.conf` to `clouseau.conf`._ Calling the sample configuration file `clouseau.conf` goes better with the older naming scheme where it was called `clouseau.ini`.

- _Remove mentions of other nodes definitions from the configuration._ Albeit the configuration file format makes it possible to specify parameters for many other nodes (to be run atop the same JVM), this has not either been verified to be working in practice or documented, and it could only create confusion for the users.  So it is better to omit discussing that in the examples for now.


README.md
```diff
@@ -19,7 +19,27 @@ However, there are some notable changes to the deployment:
...
+```
+java \
+  -server \
+  -Xmx2G \
+  -Dsun.net.inetaddr.ttl=30 \
+  -Dsun.net.inetaddr.negative.ttl=30 \
+  -XX:+ExitOnOutOfMemoryError \
+  -XX:+UseG1GC \
+  -XX:+ParallelRefProcEnabled \
+  -jar clouseau_2.13.16_3.3.0.jar \
```
poor-a, line 40: Albeit easy to find the path, it would be nice to set the relative path to the `.jar`
`clouseau/target/scala-2.13/clouseau_2.13.16_3.3.0.jar`
So we save some time for the user.

poor-a (approved):
Commented on a minor a issue. Otherwise great work.

README.md
pgj: This command is not about suggesting how to launch the application from the source tree.  It is more about giving a hint about what to do with the JAR that is released.  As such, it does not sit under the `clouseau/target/scala-2.13` directory tree.

PR #201, opened 2026-08-27
fix(`Makefile`): restrict `ci-build` to the required artifacts only

pgj:
The `ci-build` target does not need anything else but the production JAR file and the dist ZIP, so extract them from the full list of release files to build those ones only.  This may help with cutting down on the build times but it would not also require the CI system to have the documentation toolchain installed either.


poor-a (approved):
Thanks!

